PR #62, opened 2026-06-13
Load save fields with defaults so incomplete saves don't crash

### LINKED ISSUE #57: Incomplete save files crash on load: KeyError escapes the load handlers
## Summary
During a triage pass, the save-loading paths were read end-to-end. The load handlers are clearly meant to recover gracefully from a bad save (they print a warning and create a fresh object), but they only catch `(IOError, OSError, json.JSONDecodeError)` — not `KeyError`. The reader/writers access most fields with direct `[]` indexing, so a save file that is valid JSON but missing a field raises an uncaught `KeyError` and crashes the game instead of recovering.

- `src/fishE.py:227-234` (`loadPlayer`) — `except (IOError, OSError, json.JSONDecodeError) as e:` does not include `KeyError`. Same pattern in `loadStats` (`src/fishE.py:236`) and `loadTimeService` (`src/fishE.py:245`).
- `src/player/playerJsonReaderWriter.py:18-22` — direct access: `player.fishCount = playerJson["fishCount"]`, `... = playerJson["money"]`, etc. (only `energy` uses `.get(..., 100)` at line 23).
- `src/stats/statsJsonReaderWriter.py:19-24` — direct access for six fields (only `moneyLostWhileDrunk` uses `.get(..., 0)` at line 26).
- `src/world/timeServiceJsonReaderWriter.py:11-12` — `timeService.time = timeServiceJson["time"]`, `... = timeServiceJson["day"]` (no defaults).

## Why it matters
A truncated, hand-edited, or older-format save (anything missing a key) crashes on startup rather than falling back to a new game, which is exactly the failure the surrounding `try/except` was written to prevent. It also undermines the project's backward-compatibility approach to save files — the `.get(default)` pattern was added for `energy`/`moneyLostWhileDrunk` precisely so old saves keep loading, but the remaining fields don't follow it.

## Suggested fix
Make field loading backward-compatible and crash-safe. Either (a) read every persisted field with `dict.get("<field>", <default>)` consistently across the three reader/writers, or (b) add `KeyError` (and `ValueError`) to the caught exception tuples in `loadPlayer`/`loadStats`/`loadTimeService`. Option (a) is preferable since it preserves partial data instead of discarding the whole object. Add a regression test that loads a save dict missing a field and asserts graceful handling.

dmccoystephenson:
## Summary
- `createPlayerFromJson`, `createStatsFromJson`, and `createTimeServiceFromJson` now read every persisted field with `dict.get(key, <default>)`, where the default is the value the freshly-constructed model already holds. A save missing any field loads gracefully instead of raising an uncaught `KeyError`.
- This generalizes the backward-compat pattern that previously covered only `energy` and `moneyLostWhileDrunk`, so the fallback default can never drift from the model default.
- No schema changes; persisted shape is unchanged. Older/partial saves now load.

## Test plan
- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 148 passed (3 new)
- [x] New regression tests: each reader/writer loads `{}` and asserts all fields equal the model defaults (would `KeyError` before this change).

Closes #57

---

### DIFF
--- a/src/player/playerJsonReaderWriter.py
+++ b/src/player/playerJsonReaderWriter.py
@@ -14,15 +14,16 @@ class PlayerJsonReaderWriter:
         }
 
     def createPlayerFromJson(self, playerJson):
+        # Read each field with a fallback to the freshly-constructed Player's
+        # default, so a save file missing any field loads gracefully instead of
+        # raising KeyError (backwards compatibility for older/partial saves).
         player = Player()
-        player.fishCount = playerJson["fishCount"]
-        player.fishMultiplier = playerJson["fishMultiplier"]
-        player.money = playerJson["money"]
-        player.moneyInBank = playerJson["moneyInBank"]
-        player.priceForBait = playerJson["priceForBait"]
-        player.energy = playerJson.get(
-            "energy", 100
-        )  # Default to 100 for backwards compatibility
+        player.fishCount = playerJson.get("fishCount", player.fishCount)
+        player.fishMultiplier = playerJson.get("fishMultiplier", player.fishMultiplier)
+        player.money = playerJson.get("money", player.money)
+        player.moneyInBank = playerJson.get("moneyInBank", player.moneyInBank)
+        player.priceForBait = playerJson.get("priceForBait", player.priceForBait)
+        player.energy = playerJson.get("energy", player.energy)
         return player
 
     def writePlayerToFile(self, player, jsonFile):
--- a/src/stats/statsJsonReaderWriter.py
+++ b/src/stats/statsJsonReaderWriter.py
@@ -15,15 +15,27 @@ class StatsJsonReaderWriter:
         }
 
     def createStatsFromJson(self, statsJson):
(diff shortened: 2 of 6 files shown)


dmccoystephenson (commented):
Self-review rubric (adversarial):
- Scope: PASS — only the three reader/writers and their test modules; all required for #57.
- Tests-fix: PASS — each reader/writer gains a missing-all-fields test that raised KeyError before this change.
- Schema-sync: PASS — no field added/removed; persisted shape unchanged, so schemas need no edit.
- Sibling structure: PASS — all three reader/writers now use the same get(key, default) pattern (consistency improved).
- Money-format / Deterministic-tests / Headless-pygame: N/A — no money display, no randomness, no pygame paths touched.
- camelCase: PASS — production identifiers unchanged; new test names follow the existing test_ naming.
- CI: PASS — test job green.

Fix removes the direct [] access named in the issue across all three load paths; defaults are sourced from the constructed model so they cannot drift.

PR #78, opened 2026-06-13
Make tavern gambling and drinking real risk/reward choices

### LINKED ISSUE #68: Tavern gambling and drinking are pure negative-EV traps with no upside
## Summary
During a gameplay review, the two tavern activities were traced and found to be strictly losing propositions a rational player should never choose.

**Dice gambling** pays even money on a 1-in-6 chance:
- `src/location/tavern.py:148` — `self.diceThrow = random.randint(1, 6)` (six equally likely outcomes).
- `src/location/tavern.py:152` — on a win, `self.player.money += self.currentBet` (pays 1× the bet).
- `src/location/tavern.py:161` — on a loss, `self.player.money -= self.currentBet`.
- Expected value per bet = (1/6)(+bet) + (5/6)(−bet) = **−2/3 of the bet**. There is no skill component.

**Getting drunk** is strictly negative:
- `src/location/tavern.py:109` — costs `self.player.money -= 10`.
- `src/location/tavern.py:119,122` — 30% chance to additionally lose 10–50% of remaining money.
- `src/location/tavern.py:135` — burns a whole day (`increaseDay()`), with no benefit granted in return.

## Why it matters
Both options only ever cost the player, so optimal play is to ignore the tavern entirely — it is effectively dead content. Mechanics the player has no reason to use add menu noise without adding decisions.

## Suggested fix
Turn them into genuine risk/reward choices:
- **Dice:** pay out at true odds (≈5× the bet on a correct guess) so EV is roughly neutral, or add a skill/odds-selection element; surface the payout in the prompt.
- **Drinking:** grant something for the cost — e.g. restore some energy, or unlock a "rumor" (a fishing hotspot / tip) — so the money and the lost day buy a real benefit against the 30% downside.

Add/adjust tests in `tests/location/test_tavern.py` for the new payout/benefit (patching `random` for determinism, per the repo's existing pattern).

_Filed by Claude during a gameplay-balance review; claims above were verified against source._

dmccoystephenson:
## Summary
- **Dice gambling** (`src/location/tavern.py`): a correct guess now pays `DICE_WIN_MULTIPLIER` (5×) the bet instead of even money. With six equally likely faces this makes the wager roughly fair (EV ≈ 0) rather than a guaranteed long-run loss. The payout multiplier is shown in the betting prompt.
- **Getting drunk**: now has an upside as well as the existing downside — a ~30% chance to lose money *and* a ~30% chance to overhear a lucrative rumor and earn a cash tip (constants `DRUNK_LOSS_CHANCE` / `DRUNK_TIP_CHANCE`), so the $10 + lost day buys a real gamble instead of being pure loss.

## Test plan
- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — all passing
- [x] Added `test_gamble_win_pays_multiple_of_bet` (drives the real `gamble()` loop; asserts 5× payout — would fail under the old even-money logic) and `test_getDrunk_can_earn_a_tip` (tip outcome adds money). Existing `getDrunk` tests still pass (they pin `random.random` to the no-event branch).

Closes #68

---

### DIFF
--- a/src/location/tavern.py
+++ b/src/location/tavern.py
@@ -12,6 +12,18 @@ from ui.userInterface import UserInterface
 from npc.npc import NPC
 
 
+# Dice has 6 equally likely faces, so a correct guess pays 5x the bet to make
+# the wager roughly fair (EV ~= 0) instead of the old even-money payout, which
+# made gambling a guaranteed long-run loss no rational player would take.
+DICE_WIN_MULTIPLIER = 5
+
+# Chances applied while drunk, in order: lose a chunk of money, or instead pick
+# up a lucrative rumor. Giving drinking an upside (not just a downside) makes it
+# a real risk/reward choice rather than a pure money sink.
+DRUNK_LOSS_CHANCE = 0.3
+DRUNK_TIP_CHANCE = 0.3
+
+
 # @author Daniel McCoy Stephenson
 class Tavern:
     def __init__(
@@ -115,8 +127,10 @@ class Tavern:
 
         self.stats.timesGottenDrunk += 1
 
-        # Random chance of losing additional money while drunk
-        if random.random() < 0.3:  # 30% chance
+        # A night of drinking is a gamble: you might lose money in a drunken
+        # stupor, or you might overhear a lucrative rumor and come out ahead.
+        roll = random.random()
+        if roll < DRUNK_LOSS_CHANCE:
             if self.player.money > 0:
                 # Lose between 10% and 50% of remaining money
                 loss_percentage = random.uniform(0.1, 0.5)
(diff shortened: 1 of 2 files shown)


dmccoystephenson (commented):
Self-review rubric:
- Scope: PASS — tavern.py + its tests only.
- Tests-fix: PASS — test_gamble_win_pays_multiple_of_bet drives the real gamble() and asserts 5x (fails under old even-money); test_getDrunk_can_earn_a_tip covers the new upside.
- Deterministic-tests: PASS — random.random/randint patched.
- Issue resolution: PASS — dice EV is now ~neutral and drinking has a real upside.
- CI: PASS. Schema/money-format/pygame: N/A (money shown via existing int bet displays; no new float displays).

Issue #73, opened 2026-06-13
Split upgrades into distinct rod and bait paths instead of one fishMultiplier


dmccoystephenson:
## Summary
There is only one upgrade axis in the game — buying bait raises `fishMultiplier`, which is the sole way to improve fishing.

- `src/location/shop.py:108` — `self.player.fishMultiplier += 1` is the only purchasable upgrade.
- `src/location/docks.py:171` — `fishMultiplier` linearly scales yield; nothing else upgrades.
- `src/location/docks.py:152` — the catch minigame window is hardcoded `reactionTime <= 2.0`; no upgrade widens it.

## Why it matters
A single upgrade axis makes progression one-dimensional ("buy more bait"). Splitting it into a **rod** (improves catch rate / widens the minigame window) and **bait** (improves yield or rarity) gives a richer upgrade tree and lets players invest along different strategies. It also gives the minigame-window value (currently a magic number) a place to come from.

## Suggested fix
Add a `rodLevel` (or similar) to `Player`, sold separately in the shop, that widens the reaction window in `Docks.fish()` (e.g. window = base + rodLevel step) and/or raises catch rate, while bait keeps affecting yield/rarity. Persist the new field via the save system using the backward-compatible `.get(default)` pattern and add it to `schemas/player.json`. Coordinates with the `fishMultiplier` cap issue and the minigame-rework issue. Add tests under `tests/location/` for both purchase paths and their effects.

_Filed by Claude during a gameplay-design review; claims above were verified against source._

PR #82, opened 2026-06-13
Rework fishing minigame into a single quality-tiered catch

### LINKED ISSUE #74: Rework the fishing reaction minigame (unreliable, repetitive, hardcoded window)
## Summary
The fishing minigame is a wall-clock "press Enter in time" check that is unreliable in a line-buffered console, repeats every in-game hour, and has a hardcoded difficulty.

- `src/location/docks.py:143-156` — each hour prints "A fish is biting! Press Enter quickly!" and measures `reactionTime = time.time() - startTime`, succeeding only if `reactionTime <= 2.0` (a magic number).
- `src/location/docks.py:137` — this repeats `for i in range(hours)` where `hours = random.randint(1, 10)`, so a single cast can demand up to 10 timed presses interleaved with `time.sleep` calls (`:140`).
- `src/location/docks.py:157` — a `KeyboardInterrupt` during the prompt is swallowed as a miss.

## Why it matters
Reaction timing measured by wall clock in a console is inconsistent (terminal/input buffering, latency), and repeating it up to 10 times per cast with `sleep` pauses makes a routine action tedious. The fixed 2.0s threshold also gives upgrades nothing to improve.

## Suggested fix
Make the catch mechanic more forgiving and meaningful: e.g. resolve the whole cast with one interaction (or a short reel-in sequence) rather than per-hour, derive the success window from a rod/bait upgrade instead of a hardcoded `2.0` (see the rod/bait split issue), and map reaction speed to a quality/rarity tier rather than pass/fail. Keep the timed portion mockable so it can be tested (the existing tests patch I/O and `random`); add coverage under `tests/location/test_docks.py`.

_Filed by Claude during a gameplay-design review; claims above were verified against source._

dmccoystephenson:
## Summary
`src/location/docks.py` `fish()`:
- **One timing challenge per cast** instead of a pass/fail repeated every in-game hour (previously up to 10 timed `input()` prompts with `time.sleep` between each).
- Reaction speed maps to a **catch-quality tier** — perfect (≤ half the window), solid (≤ window), or poor (miss/over) — which scales the yield, rather than binary success counting.
- The window is the rod-derived `reactionWindow` from #73, so the hardcoded 2.0s is gone and a better rod genuinely helps.
- Hours still pass and consume energy (10/hour) in a now-silent loop.

## Test plan
- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 161 passed
- [x] Replaced the old binary `test_fish_interactive_failure` (which asserted exactly 1 fish on a miss — no longer the model) with `test_fish_slow_reaction_yields_fewer_than_fast` (slow ≥ 1 but < fast, identical rolls). Existing energy/time/limited-energy and rod-window tests still hold.

Closes #74

---

### DIFF
--- a/src/location/docks.py
+++ b/src/location/docks.py
@@ -154,67 +154,46 @@ class Docks:
                 self.currentPrompt.text = "You're too tired to fish! Go home and sleep."
                 return
 
-        successfulCatches = 0
-        totalAttempts = 0
-
         # A better rod widens the timing window, making catches more forgiving.
         reactionWindow = REACTION_BASE_WINDOW + (self.player.rodLevel - 1) * ROD_WINDOW_STEP
 
+        # One timing challenge per cast (not a pass/fail repeated every hour):
+        # how quickly you set the hook maps to a catch-quality tier.
+        print("Cast your line... press Enter the moment you feel a bite! ")
+        sys.stdout.flush()
+        startTime = time.time()
+        try:
+            input()
+            reactionTime = time.time() - startTime
+        except (KeyboardInterrupt, EOFError):
+            reactionTime = reactionWindow + 1.0  # an aborted input counts as a miss
+
+        if reactionTime <= reactionWindow / 2:
+            quality, qualityLabel = 1.0, "A perfect hook!"
+        elif reactionTime <= reactionWindow:
+            quality, qualityLabel = 0.6, "A solid hook."
+        else:
+            quality, qualityLabel = 0.25, "The fish nearly got away."
+
+        # Spend the fishing hours: time passes and energy is consumed.
         for i in range(hours):
-            print("><> ")
-            sys.stdout.flush()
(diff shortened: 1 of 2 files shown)


dmccoystephenson (commented):
Self-review rubric:
- Scope: PASS — docks.py fish() + its test only.
- Tests-fix: PASS — the obsolete binary 'exactly 1 on miss' test was replaced (not weakened) with a gradient test reflecting the new quality model; energy/time/rod-window tests still hold.
- Issue resolution: PASS — single challenge (no per-hour repetition/sleeps), quality tiers, rod-derived window (no hardcoded 2.0s).
- Deterministic-tests: PASS — time.time/input/randint patched.
- CI: PASS. Schema/money-format/pygame: N/A.

PR #87, opened 2026-06-13
Route in-game input through the UI so pygame needs no console

dmccoystephenson:
## Why
#42: under the pygame front-end the player still had to drop to the console for input. Even after the UI now has `showOptions`/`showDialogue`, the game called bare `input()` for bank amounts, the tavern bet, the fishing reaction, and the stats "continue" — none of which work in the pygame window.

## What changed
- **`BaseUserInterface`** gains the input primitives: abstract `promptForText(promptText)` and `timedKeyPress(message)`, plus a concrete `promptForNumber(promptText)` built on `promptForText`.
- **Console (`UserInterface`)** and **`PygameUserInterface`** both implement them — pygame captures on-screen text entry (type + Enter/Backspace) and reaction timing in its event loop.
- In-game input routed through the UI:
  - `Bank.deposit/withdraw` → `promptForNumber`
  - `Tavern.changeBet` → `promptForNumber`
  - `Docks.fish` reaction minigame → `timedKeyPress`
  - `Home.displayStats` → builds a text block and renders via `showDialogue`
- No bare `input()` remains in gameplay code.

## Scope / follow-up
The pre-game **save-slot manager** (`_selectSaveFile`/`_deleteSaveFile`) still uses the console: it runs in `__init__` before the UI (and player/timeService) exist and renders custom multi-line listings, so routing it cleanly needs UI-before-state construction + output rendering. Tracked as a follow-up (linked below). This PR therefore **advances** #42 (all in-play input) rather than closing it.

## Test plan
- [x] `python3 -m compileall -q src` clean
- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 186 passed
- [x] New tests: `promptForNumber` parsing (base), console `promptForText`/`promptForNumber`/`timedKeyPress`; bank/tavern/docks/home tests updated to drive the UI primitives (no more bare-`input` patching). Behavior preserved on the console.

Advances #42.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

_drafted by Claude on behalf of Daniel Stephenson_

PR #89, opened 2026-06-13
Route the save-file manager through the UI (closes #42)

### LINKED ISSUE #42: Support all needed input through pygame UI
When using the pygame UI, the user has to switch back to the console to provide input sometimes. All input should be capable of being handled by the pygame UI so the user doesn't have to context switch.

### LINKED ISSUE #88: Route the pre-game save-slot manager through the UI (last console-only input)
## Summary
After PR #87, all in-play input goes through the UI (so pygame needs no console mid-game). The one remaining console-only input is the **save-slot manager** shown at launch.

- `src/fishE.py` `_selectSaveFile` (`:140`) and `_deleteSaveFile` (`:184`, `:191`, `:195`) use bare `input()` and bare `print()` for a custom multi-line slot listing.
- It runs inside `FishE.__init__` **before** the UI (and `player`/`timeService`) are constructed, so it can't simply call `self.userInterface`.

## Why it matters
This is the final place a pygame player must use the console (to pick/delete a save), so #42 isn't fully closed until it's handled.

## Suggested approach
Construct the front-end earlier (it only needs the input/continue primitives for this menu, not the gameplay header), then render the slot list via `showOptions`/`showDialogue` and read choices via `promptForText`. The `promptForText`/`showDialogue` primitives added in #87 already support both console and pygame; the main work is the construction-order change and converting the custom `print` listing to UI rendering.

_Filed by Claude as the tracked remainder of #42._

dmccoystephenson:
## Why
This is the last piece of #42 (tracked as #88). The pre-game save-slot manager used bare `input()`/`print()`, so a pygame player still had to use the console to pick or delete a save.

## What changed
- **`FishE.__init__`** now builds the UI **early** (with default new-game state) before the save-file manager runs, then loads the chosen slot over those defaults and points the UI at the loaded state. (This also simplified the load block — defaults are created once instead of via `else` branches.)
- **`_selectSaveFile` / `_deleteSaveFile`** present save slots and actions (Load / Create New / Delete / Quit; and the delete confirmation) as numbered `showOptions` choices, and use `showDialogue` for messages — so the whole flow renders and reads input through the active front-end.

`grep` confirms **no bare `input()` remains anywhere** in the game; every input goes through the UI. Together with #87 (in-play input), #42 is fully resolved.

## Test plan
- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 190 passed
- [x] New `test_fishE` cases for the now-testable menu: new-game selects the next slot, loading selects the chosen slot, delete confirmed vs cancelled. Existing `test_initialization` still passes with the construction reorder.

Closes #42
Closes #88

---

### DIFF
--- a/src/fishE.py
+++ b/src/fishE.py
@@ -35,38 +35,39 @@ class FishE:
         self.statsJsonReaderWriter = StatsJsonReaderWriter()
         self.saveFileManager = SaveFileManager()
 
-        # Migrate old save files to new format if they exist
-        self.saveFileManager.migrate_old_save_files()
-
-        # Show save file selection menu
-        self._selectSaveFile()
-
-        # if save file exists, load it
-        player_path = self.saveFileManager.get_save_path("player.json")
-        if os.path.exists(player_path) and os.path.getsize(player_path) > 0:
-            self.loadPlayer()
-        else:
-            self.player = Player()
-
-        # if save file exists, load it
-        stats_path = self.saveFileManager.get_save_path("stats.json")
-        if os.path.exists(stats_path) and os.path.getsize(stats_path) > 0:
-            self.loadStats()
-        else:
-            self.stats = Stats()
-
-        # if save file exists, load it
-        time_path = self.saveFileManager.get_save_path("timeService.json")
-        if os.path.exists(time_path) and os.path.getsize(time_path) > 0:
-            self.loadTimeService()
-        else:
-            self.timeService = TimeService(self.player, self.stats)
-
+        # Start from default (new-game) state, then build the UI so the save-file
+        # manager can render and read input through the active front-end. A
+        # chosen save is loaded over these defaults below.
+        self.player = Player()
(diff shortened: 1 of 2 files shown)


dmccoystephenson (commented):
Self-review:
- Scope: PASS — fishE __init__ reorder + save-menu rewrite + tests.
- Issue: PASS — no bare input() remains anywhere; save menu now renders/reads via showOptions/showDialogue, so pygame needs no console. Closes #42/#88.
- Behavior: PASS — console still works (slots/actions as numbered options); load-if-exists logic preserved (defaults created once instead of else-branches).
- Tests: PASS — 190; new menu tests (new/load/delete-confirmed/cancelled); test_initialization validates the construction reorder.
- Persistence safety: load/save methods, saveFileManager, schemas, and reader/writers untouched.
- CI: PASS.

PR #113, opened 2026-07-18
Add container support: configurable web-UI bind address + Dockerfile

dmccoystephenson:
## Summary

Two changes needed to run FishE's web front-end in a Docker container behind a reverse proxy (Traefik), for deployment in `Stephenson-Software/gateway`:

- `WebUserInterface.__init__` already accepts `host`/`port` params, but they were never plumbed up through `UserInterfaceFactory.create_user_interface`'s `WEB` branch, which called it with no override — always binding the hardcoded default `127.0.0.1:8000`. Inside a container on a bridge network, a process bound to `127.0.0.1` is unreachable from a sibling container (e.g. a reverse proxy), so this makes the `WEB` branch read `FISHE_WEB_HOST`/`FISHE_WEB_PORT` env vars, defaulting to the existing `127.0.0.1:8000` so behavior is unchanged for anyone not setting them.
- Adds a `Dockerfile` (`python:3.12-slim`, stdlib-only — no pygame, matching the web front-end's own "no extra dependencies" footprint) that runs `examples/web_app.py` with `FISHE_WEB_HOST=0.0.0.0` baked in via `ENV`, and creates `/app/data` (the `SaveFileManager` default save directory, resolved relative to the process cwd) owned by a non-root user so a volume mounted over it stays writable.

`examples/web_app.py`'s startup print now reflects the actual host/port instead of a hardcoded string.

## Test plan

- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest --verbose -vv --cov=src --cov-report=term-missing --cov-report=xml:cov.xml` — 327 passed, `userInterfaceFactory.py` at 100% coverage
- [x] `black`/`autoflake` run locally; only the files actually touched by this change are included in the diff (pre-existing formatting drift elsewhere in the repo was reverted, out of scope for this PR)
- [x] Manual smoke test: ran `FishE(interfaceType=UIType.WEB)` with `FISHE_WEB_HOST=0.0.0.0 FISHE_WEB_PORT=8123` and confirmed the server bound `0.0.0.0` and `GET /state` returned 200 with the expected JSON

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,30 @@
+FROM python:3.12-slim
+
+# No third-party deps: the game (and its web front-end) is stdlib-only —
+# pygame is only imported lazily by the pygame front-end, which this image
+# never selects, so it's deliberately not installed here. Matches the
+# gateway-dashboard/fleet-dashboard services' "no wheel supply chain to keep
+# patched" approach.
+WORKDIR /app
+COPY src/ ./src/
+COPY examples/ ./examples/
+
+# Create the data/ directory (see saveFileManager.py's default
+# data_directory="data", resolved relative to the process cwd -> /app/data
+# here) and hand it to a non-root user before switching to it, so a named
+# volume mounted over /app/data on first run inherits writable ownership.
+RUN useradd --system --no-create-home fishe \
+    && mkdir -p /app/data \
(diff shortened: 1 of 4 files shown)


dmccoystephenson:
@copilot address comments

PR #115, opened 2026-07-19
Validate loaded save data against schemas; wire Config into Player/SaveFileManager

### LINKED ISSUE #109: Wire the existing Config class into Player and SaveFileManager instead of hardcoding defaults
## Summary
`src/config/config.py` defines a `Config` class holding every save-file path and every initial player value, and it has its own dedicated test (`tests/config/test_config.py`), but nothing in the game ever instantiates or reads it — every value it defines is duplicated as a hardcoded literal elsewhere instead.

- `src/config/config.py:1-16` — `Config.__init__` sets `dataDirectory = "data"`, `playerSaveFile`, `statsSaveFile`, `timeServiceSaveFile`, and initial values `initialMoney = 20`, `initialEnergy = 100`, `initialFishCount = 0`, `initialMoneyInBank = 0.01`, `initialFishMultiplier = 1`, `initialPriceForBait = 50`.
- `grep -rn "Config" --include=*.py src/ examples/ tests/` (excluding `tests/config/`) returns **only the class definition itself** — `Config` is never imported or constructed anywhere else in `src/` or `examples/`.
- `src/player/player.py:9-16` hardcodes the same six values independently (`self.money = 20`, `self.priceForBait = 50`, etc.) rather than reading them from `Config`.
- `src/saveFileManager.py:11` — `def __init__(self, data_directory="data")` duplicates `Config.dataDirectory` as a separate string literal.
- `src/fishE.py:37` — `self.saveFileManager = SaveFileManager()` constructs it with no `data_directory` argument, so even the one constructor parameter that *could* take a configured value never receives one.

## Why it matters
The project already has a working, tested precedent for runtime-configurable defaults: `Player.operatorMode` (`src/player/player.py:43-47`) reads `FISHE_OPERATOR_MODE` from the environment so a whole session's behavior can be changed without editing code. `Config` was clearly built to play the same role for starting stats and save location, but is entirely disconnected from the game it configures — anyone who edits `config.py` expecting it to change a new game's starting money, or expecting `dataDirectory` to relocate saves, will find it does nothing.

## Suggested fix
Construct a `Config()` in `FishE.__init__` (`src/fishE.py`) and thread it through: pass `config.dataDirectory` into `SaveFileManager(data_directory=...)` (`src/fishE.py:37`), and give `Player.__init__` (`src/player/player.py`) an optional `config: Config = None` parameter that seeds `money`/`energy`/`fishCount`/`moneyInBank`/`fishMultiplier`/`priceForBait` from it, defaulting to today's literals when omitted so existing callers and tests are unaffected. This turns `Config` from dead code into the same kind of session-level override `operatorMode` already provides (e.g. a future `FISHE_DATA_DIR` env var reusing the same pattern). Add a test to `tests/player/test_player.py` and `tests/test_saveFileManager.py` constructing with a custom `Config`/`data_directory` and asserting it takes effect.

_Filed by Claude during an automated feature-opportunity triage pass; claims above were verified against source._

### LINKED ISSUE #111: Validate loaded save data against the existing JSON schemas instead of accepting any well-formed JSON
## Summary
The repo ships formal JSON Schemas describing exactly what a valid save file looks like, including value ranges, but nothing in the game ever reads or validates against them — any well-formed JSON that merely has the right key names is accepted as-is, regardless of whether the values make sense.

- `schemas/player.json` — declares real constraints beyond "does this key exist," e.g. `"energy": {"type": "integer", "minimum": 0, "maximum": 200}`, `"homeTier": {"type": "integer", "minimum": 0, "maximum": 5}`, `"rentalProperties": {"items": {"type": "integer", "minimum": 1, "maximum": 3}}`, and a `"required"` list of six fields.
- `schemas/stats.json` and `schemas/timeService.json` similarly describe every persisted field (e.g. `timeService.json`'s `"time": {"minimum": 0, "maximum": 23}`).
- `grep -rn "jsonschema\|schemas/" --include=*.py src/ tests/` returns no results — no Python module imports the `jsonschema` package or opens anything under `schemas/`. The schema files are referenced by nothing.
- `src/player/playerJsonReaderWriter.py:24-45` (`createPlayerFromJson`) loads every field with `dict.get(key, default)` — this recovers from *missing* keys (the fix for #57) but performs no type or range check, so a save with e.g. `"energy": -500` or `"homeTier": 99` loads without complaint and is then handled by game logic that assumes in-range values (e.g. `src/housing/housing.py:62-65`'s `tierInfo` raises `ValueError` for an out-of-range tier, uncaught anywhere in the load path).

## Why it matters
The schemas encode real constraints the team has already thought through (energy caps, valid tier ranges, valid property-type ids) but that knowledge is never applied at runtime. A hand-edited or partially-corrupted-but-syntactically-valid save (a real scenario the project already cares about — see #57 and #56) can load successfully and then crash or misbehave later, deeper in game logic, at a point that's harder to trace back to a bad save than a clear validation error at load time would be.

## Suggested fix
Add a small `validate_against_schema(data, schema_path)` helper (using the `jsonschema` package, or a minimal hand-rolled range/type checker to avoid a new dependency if the project prefers staying stdlib-only per the README's "no extra dependencies" web-UI claim) and call it from `PlayerJsonReaderWriter.createPlayerFromJson`, `StatsJsonReaderWriter`, and `TimeServiceJsonReaderWriter` before applying the loaded values, alongside their existing `.get()`-with-defaults pattern. On a validation failure, fall back the same way `FishE.loadPlayer`/`loadStats`/`loadTimeService` (`src/fishE.py:271-308`) already fall back on `IOError`/`JSONDecodeError` — a fresh default object plus the existing warning print — rather than adding a new failure mode. Add tests loading a save with an out-of-range value (e.g. `energy: -1` or `homeTier: 99`) and asserting it falls back cleanly instead of propagating a `ValueError` from deep in `housing.py`/`business.py`.

_Filed by Claude during an automated feature-opportunity triage pass; claims above were verified against source._

dmccoystephenson:
## Summary
- Add a `validate_against_schema()` helper (`src/validation/schemaValidator.py`) and call it from each `*JsonReaderWriter`'s `create*FromJson`, after the existing `.get()`-with-defaults fallback runs — so a save missing keys still loads fine (backwards compatibility is untouched), but an out-of-range or wrong-type value that *was* present (e.g. `energy: -500`, `homeTier: 99`, a stringified stat) is now rejected with a `jsonschema.exceptions.ValidationError` instead of silently loading and surfacing later as an uncaught `ValueError` deep in `housing.py`/`business.py`.
- Catch that `ValidationError` in `FishE.loadPlayer`/`loadStats`/`loadTimeService` the same way `IOError`/`JSONDecodeError` are already caught — falls back to a fresh default object plus the existing warning print, no new failure mode.
- Construct a `Config()` in `FishE.__init__` and thread it through: `SaveFileManager(data_directory=config.dataDirectory)`, and an optional `Player(config=None)` parameter that seeds `money`/`fishCount`/`moneyInBank`/`fishMultiplier`/`priceForBait` from it (defaulting to today's literals when omitted, so every existing caller/test is unaffected). This turns `Config` from dead code into an actual session-level override, matching the pattern `Player.operatorMode` already uses for `FISHE_OPERATOR_MODE`.
  - `energy` is deliberately **not** sourced from `Config.initialEnergy` — the housing ladder (`housing.HOUSING_TIERS[0]["maxEnergy"]`, currently 60) is the real source of truth for starting energy, and `Config.initialEnergy` (100) predates/ignores that and would let a configured player start above the Homeless cap. See the comment in `Player.__init__`.

Closes #111
Closes #109

## Test plan
- [x] Full suite green: `pytest --verbose -vv --cov=src --cov-report=term-missing --cov-report=xml:cov.xml` → 345 passed, 100% coverage on all newly-touched modules (`schemaValidator.py`, the three `*JsonReaderWriter`s)
- [x] Backend-only change — no front-end-visible menu/dialogue, so no console/pygame/web parity work needed
- [x] New tests: out-of-range/wrong-type values raise `ValidationError` in each reader/writer; `FishE.loadPlayer` recovers gracefully from an out-of-range save; `Player(config)` seeds from a custom `Config` while `Player()` is unaffected; `FishE.__init__` wires `self.config` into `SaveFileManager` and `Player`
- [x] Self-review performed per project convention (no configured reviewer)

---

### DIFF
--- a/src/fishE.py
+++ b/src/fishE.py
@@ -1,5 +1,6 @@
 import os
 import json
+from jsonschema.exceptions import ValidationError
 from location import bank, docks, home, shop, tavern
 from location.enum.locationType import LocationType
 from player.player import Player
@@ -15,6 +16,7 @@ from saveFileManager import SaveFileManager
 from achievements import achievements
 from achievements.achievements import GOAL_AMOUNT, GOAL_MILESTONE_NAME
 from housing import housing
+from config.config import Config
 
 # Which front-end the game runs. Swap to UIType.PYGAME (or a future web type)
 # here to change the interface — the rest of the game is front-end agnostic.
@@ -26,15 +28,16 @@ class FishE:
     def __init__(self, interfaceType=INTERFACE_TYPE):
         self.running = True
 
+        self.config = Config()
         self.playerJsonReaderWriter = PlayerJsonReaderWriter()
         self.timeServiceJsonReaderWriter = TimeServiceJsonReaderWriter()
         self.statsJsonReaderWriter = StatsJsonReaderWriter()
-        self.saveFileManager = SaveFileManager()
+        self.saveFileManager = SaveFileManager(data_directory=self.config.dataDirectory)
 
         # Start from default (new-game) state, then build the UI so the save-file
         # manager can render and read input through the active front-end. A
         # chosen save is loaded over these defaults below.
-        self.player = Player()
+        self.player = Player(self.config)
         self.stats = Stats()
         self.timeService = TimeService(self.player, self.stats)
         self.prompt = Prompt("What would you like to do?")
@@ -271,10 +274,10 @@ class FishE:
                 self.player = self.playerJsonReaderWriter.readPlayerFromFile(
                     playerSaveFile
                 )
-        except (IOError, OSError, json.JSONDecodeError) as e:
+        except (IOError, OSError, json.JSONDecodeError, ValidationError) as e:
             print(f"\n Warning: Failed to load player data: {e}")
             print(" Creating new player...")
-            self.player = Player()
+            self.player = Player(self.config)
 
     def loadStats(self):
         try:
@@ -282,7 +285,7 @@ class FishE:
                 self.saveFileManager.get_save_path("stats.json"), "r"
             ) as statsSaveFile:
                 self.stats = self.statsJsonReaderWriter.readStatsFromFile(statsSaveFile)
-        except (IOError, OSError, json.JSONDecodeError) as e:
+        except (IOError, OSError, json.JSONDecodeError, ValidationError) as e:
             print(f"\n Warning: Failed to load stats data: {e}")
             print(" Creating new stats...")
             self.stats = Stats()
@@ -297,7 +300,7 @@ class FishE:
                         timeServiceSaveFile, self.player, self.stats
                     )
                 )
-        except (IOError, OSError, json.JSONDecodeError) as e:
+        except (IOError, OSError, json.JSONDecodeError, ValidationError) as e:
             print(f"\n Warning: Failed to load time service data: {e}")
             print(" Creating new time service...")
             self.timeService = TimeService(self.player, self.stats)
--- a/src/player/player.py
+++ b/src/player/player.py
@@ -5,14 +5,17 @@ from housing import housing
 
 # @author Daniel McCoy Stephenson
 class Player:
-    def __init__(self):
-        self.fishCount = 0
-        self.money = 20
-        self.moneyInBank = 0.01
-        self.fishMultiplier = 1
-        self.priceForBait = 50
+    def __init__(self, config=None):
+        self.fishCount = 0 if config is None else config.initialFishCount
+        self.money = 20 if config is None else config.initialMoney
+        self.moneyInBank = 0.01 if config is None else config.initialMoneyInBank
+        self.fishMultiplier = 1 if config is None else config.initialFishMultiplier
+        self.priceForBait = 50 if config is None else config.initialPriceForBait
         # Starts at the Homeless tier's energy cap (see src/housing) - a
-        # fresh player hasn't found anywhere to stay yet.
+        # fresh player hasn't found anywhere to stay yet. Deliberately not
+        # sourced from Config.initialEnergy: the housing ladder is the
+        # source of truth for energy caps (see src/housing/housing.py), and
+        # a flat configured starting energy could exceed the Homeless cap.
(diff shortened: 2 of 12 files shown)


dmccoystephenson:
{
  "body": "Self-review: no issues found. Verified backwards compatibility (missing-key saves still load via the existing .get()-with-defaults fallback; validation only runs against the reconstructed object, not the raw input, so it cannot reject a save merely for omitting optional keys). Verified Config.initialEnergy (100) intentionally is not wired into Player, since it would let a configured player exceed the Homeless housing tier's real energy cap (60) - see the comment in Player.__init__. Confirmed this is backend-only (no front-end-visible menu/dialogue), so no console/pygame/web parity work is needed. Full suite green (345 passed) with 100% coverage on every newly-touched module.",
  "event": "COMMENT",
  "comments": []
}

dmccoystephenson:
Self-review: no issues found. Verified backwards compatibility (missing-key saves still load via the existing .get()-with-defaults fallback; validation only runs against the reconstructed object, not the raw input, so it cannot reject a save merely for omitting optional keys). Verified Config.initialEnergy (100) intentionally is not wired into Player, since it would let a configured player exceed the Homeless housing tier's real energy cap (60) - see the comment in Player.__init__. Confirmed this is backend-only (no front-end-visible menu/dialogue), so no console/pygame/web parity work is needed. Full suite green (345 passed) with 100% coverage on every newly-touched module.

---

dmccoystephenson:
(Ignore the raw-JSON comment above - it was an artifact from an earlier attempt to post this same self-review through the reviews API; the review content is in the comment below it.)

PR #123, opened 2026-07-26
Rebind TimeService to the loaded player and stats after save load

### LINKED ISSUE #120: TimeService keeps stale player/stats references when a save slot has no timeService.json
## Problem

`FishE.__init__` builds `TimeService(self.player, self.stats)` from the new-game defaults (`src/fishE.py:40-42`), then loads each save file **only if that file exists and is non-empty** (`src/fishE.py:55-65`). `loadPlayer()` and `loadStats()` rebind `self.player` / `self.stats` to brand-new objects, but nothing rebinds the references `TimeService` already captured. Only `loadTimeService()` fixes this, by reconstructing `TimeService` with the current objects (`src/fishE.py:298-302`).

So if `player.json` (or `stats.json`) exists but `timeService.json` does not, `self.timeService.player` and `self.timeService.stats` keep pointing at the discarded default objects, while the rest of the game uses the loaded ones.

Everything on the daily tick runs off those stale references — `TimeService.increaseDay` (`src/world/timeService.py:46-68`) applies bank interest to `self.player`, then hands `self.player, self.stats` to `business.runDailyProduction`, `investments.runDailyIncome` and `housing.runDailyRent`. With a stale reference, a new day silently credits interest, crew catch and rental income to an object nobody reads, and charges rent against it too — so the loaded player never earns or is billed anything on a day rollover.

## How it is reachable

`SaveFileManager.migrate_old_save_files` (`src/saveFileManager.py:136-161`) gates on `player.json` only, then moves whichever of the three files happen to exist:

```python
if not os.path.exists(old_player):
    return False
...
if os.path.exists(old_time):
    shutil.move(old_time, os.path.join(slot_1_path, "timeService.json"))
```

An old save directory holding `player.json` but no `timeService.json` therefore migrates into a slot with exactly that shape. The same shape also results from a save interrupted partway through `FishE.save()` (it writes the three files in sequence, `player.json` first), or from a `timeService.json` truncated to zero bytes — the `os.path.getsize(...) > 0` guard skips it and leaves the stale references in place.

## Expected

After the load block, `self.timeService.player` and `self.timeService.stats` are the same objects the rest of the game uses, whichever subset of save files was present.

## Suggested fix

Rebind them once after the loads, alongside the existing `self.userInterface.player` / `.timeService` repointing at `src/fishE.py:67-69`, and cover it with a test that loads a slot containing only `player.json` and asserts `game.timeService.player is game.player`.

dmccoystephenson:
## Summary

- `FishE.__init__` builds `TimeService(self.player, self.stats)` from the new-game defaults, then loads each save file only if it exists and is non-empty. `loadPlayer()`/`loadStats()` rebind `self.player`/`self.stats` to brand-new objects, but only `loadTimeService()` rebuilds the `TimeService` around them — so a slot holding `player.json` but no `timeService.json` left `self.timeService.player`/`.stats` pointing at the discarded defaults.
- Everything on the daily tick runs off those references: `TimeService.increaseDay` credits bank interest to `self.player`, then hands `self.player, self.stats` to `business.runDailyProduction`, `investments.runDailyIncome` and `housing.runDailyRent`. With a stale reference, a day rollover silently credited interest, crew catch and rental income to an object nobody reads, and charged rent against it too.
- Fixed by rebinding `self.timeService.player`/`.stats` once after the load block, alongside the existing `self.userInterface.player`/`.timeService` repointing.

That slot shape is reachable in practice: `SaveFileManager.migrate_old_save_files` gates on `player.json` only and moves whichever of the three files exist; the same shape also results from a save interrupted partway through `FishE.save()` (which writes `player.json` first), or from a zero-byte `timeService.json` that the `getsize(...) > 0` guard skips.

## Test plan

- [x] `python3 -m pytest --cov=src --cov-report=term-missing --cov-report=xml:cov.xml` — 371 passed (368 before), coverage 94%
- [x] All three new tests were confirmed to fail against `main` (stashing only the `src/fishE.py` change) and pass with it
- [x] No front-end change — this is game-state wiring in `FishE.__init__`, below the `BaseUserInterface` contract, so console/pygame/web are all unaffected
- [x] No `Player`/`Stats`/`TimeService` field added, renamed or retyped, so `schemas/*.json` and the `*JsonReaderWriter` classes are unchanged (re-verified all three still match their reader-writer field-for-field)

New tests in `tests/test_fishE.py`, driving the real `FishE.__init__` against a temp save slot with only the save-slot menu and front-end stubbed:

- `test_init_rebinds_timeService_to_loaded_player_without_timeService_file` — slot with `player.json` + `stats.json`, no `timeService.json`
- `test_init_rebinds_timeService_when_only_player_file_present` — interrupted-save shape
- `test_init_daily_tick_credits_the_loaded_player` — behavioural: `increaseDay()` reaches the player the game actually reads

Closes #120

---

### DIFF
--- a/src/fishE.py
+++ b/src/fishE.py
@@ -64,6 +64,15 @@ class FishE:
         if os.path.exists(time_path) and os.path.getsize(time_path) > 0:
             self.loadTimeService()
 
+        # loadPlayer()/loadStats() rebind self.player/self.stats to brand-new
+        # objects, but only loadTimeService() rebuilds the TimeService around
+        # them. When a slot has player.json/stats.json but no timeService.json,
+        # the TimeService built from the defaults above would keep pointing at
+        # the discarded objects, so every daily tick (interest, crew catch,
+        # investment income, rent) would apply to a player nobody reads.
+        self.timeService.player = self.player
+        self.timeService.stats = self.stats
+
         # Point the UI at the (possibly reloaded) game state.
         self.userInterface.player = self.player
         self.userInterface.timeService = self.timeService
--- a/tests/test_fishE.py
+++ b/tests/test_fishE.py
@@ -110,6 +110,91 @@ def createGameForPersistence(data_directory):
     return game
 
 
+def createGameThroughInit(data_directory, saveFiles):
+    # Run the real FishE.__init__ against a temp save slot holding exactly
+    # saveFiles ({filename: json-serializable}), with only the save-slot menu
+    # and the front-end stubbed out - everything else is the real wiring, so
+    # the load block and the state it hands to TimeService are exercised.
+    fishE.Player = Player
+    fishE.Stats = Stats
+    fishE.TimeService = TimeService
+    fishE.Prompt = Prompt
+    fishE.PlayerJsonReaderWriter = PlayerJsonReaderWriter
+    fishE.StatsJsonReaderWriter = StatsJsonReaderWriter
+    fishE.TimeServiceJsonReaderWriter = TimeServiceJsonReaderWriter
+    fishE.SaveFileManager = SaveFileManager
+
+    slot = os.path.join(data_directory, "slot_1")
+    os.makedirs(slot, exist_ok=True)
+    for filename, contents in saveFiles.items():
+        with open(os.path.join(slot, filename), "w") as f:
+            json.dump(contents, f)
+
+    config = Config()
+    config.dataDirectory = data_directory
+
+    def selectSlotOne(self):
+        self.saveFileManager.select_save_slot(1)
+
+    with patch.object(fishE, "Config", return_value=config), patch.object(
+        fishE, "UserInterfaceFactory", MagicMock()
+    ), patch.object(fishE.FishE, "_selectSaveFile", selectSlotOne):
+        return fishE.FishE()
+
+
+def test_init_rebinds_timeService_to_loaded_player_without_timeService_file():
+    with tempfile.TemporaryDirectory() as data_directory:
+        # prepare/call - a slot holding player.json and stats.json but no
+        # timeService.json, which is the shape migrate_old_save_files() produces
+        # from an old save that never had one
+        game = createGameThroughInit(
+            data_directory,
+            {
+                "player.json": PlayerJsonReaderWriter().createJsonFromPlayer(Player()),
+                "stats.json": StatsJsonReaderWriter().createJsonFromStats(Stats()),
+            },
+        )
+
+        # check - the TimeService drives the same objects the rest of the game
+        # uses, so daily interest/income/rent land on the loaded player
+        assert game.timeService.player is game.player
+        assert game.timeService.stats is game.stats
+
+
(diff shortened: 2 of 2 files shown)


src/fishE.py
```diff
@@ -64,6 +64,15 @@ def __init__(self, interfaceType=INTERFACE_TYPE):
         if os.path.exists(time_path) and os.path.getsize(time_path) > 0:
             self.loadTimeService()
 
+        # loadPlayer()/loadStats() rebind self.player/self.stats to brand-new
+        # objects, but only loadTimeService() rebuilds the TimeService around
+        # them. When a slot has player.json/stats.json but no timeService.json,
+        # the TimeService built from the defaults above would keep pointing at
+        # the discarded objects, so every daily tick (interest, crew catch,
+        # investment income, rent) would apply to a player nobody reads.
+        self.timeService.player = self.player
```
dmccoystephenson, line 73: Rebinding attributes here keeps the fix small, but it does leave two places that must agree about what the current player is (this block and the `TimeService.__init__` that captured them). A more structural fix would be to defer constructing `TimeService` until after the load block entirely — it is only needed earlier because `UserInterfaceFactory.create_user_interface` takes it at line 44 so the save-file menu can render. Not worth the churn in this PR, and the comment above documents the coupling, but worth knowing if the load sequence is reworked later.

tests/test_fishE.py
```diff
@@ -110,6 +110,91 @@ def createGameForPersistence(data_directory):
     return game
 
 
+def createGameThroughInit(data_directory, saveFiles):
+    # Run the real FishE.__init__ against a temp save slot holding exactly
+    # saveFiles ({filename: json-serializable}), with only the save-slot menu
+    # and the front-end stubbed out - everything else is the real wiring, so
+    # the load block and the state it hands to TimeService are exercised.
+    fishE.Player = Player
```
dmccoystephenson, line 118: This helper reassigns `fishE`'s module globals without restoring them, which is the same thing `createFishE` (line 22) and the existing persistence tests already do in this file, so it is consistent rather than new. It is also order-independent in both directions: `createFishE` re-mocks every global it cares about, and this helper re-binds every real class it needs, so neither can be poisoned by the other regardless of which runs first. Confirmed by running the full suite (371 passed).

dmccoystephenson (commented):
Self-review: the fix is correct and minimally scoped. Verified `loadPlayer`/`loadStats`/`loadTimeService` are called only from `__init__` (src/fishE.py:57, :61, :65 — no other call sites), so a single rebind after the load block covers every path that can replace `self.player`/`self.stats`, including the `except` fallbacks inside the loaders. It runs unconditionally, so the already-correct case (timeService.json present) re-assigns the same objects and is a no-op. Placement is before the `self.locations` dict is built, so every location gets the same TimeService instance it always did. Schemas and reader-writers re-checked field-for-field and are untouched by this change; no front-end path is involved. Two notes below, neither blocking.

PR #128, opened 2026-07-27
Reject sub-$1 tavern bets and fix stale gambling payout dialogue

### LINKED ISSUE #126: Old Tom's tavern dialogue says gambling will "double your money", but the dice pay out 5x
## Problem

One of Old Tom the Barkeep's canned dialogue answers still describes the old even-money dice payout:

`src/location/tavern.py:71`
```
"Or if you're feeling lucky, you can gamble with the dice. Place a bet, pick a number from 1 to 6, "
"and if the dice matches your choice, you'll double your money!"
```

The actual payout is 5x the bet:

- `src/location/tavern.py:20` — `DICE_WIN_MULTIPLIER = 5`, with a comment explaining the change away from "the old even-money payout, which made gambling a guaranteed long-run loss".
- `src/location/tavern.py:211` — `winAmount = self.currentBet * DICE_WIN_MULTIPLIER`
- The gamble menu's own descriptor (`src/location/tavern.py:200-202`) already says "he pays out %dx your bet" with `DICE_WIN_MULTIPLIER` interpolated.

So the one in-game character whose job is to explain the mechanic tells the player something the rest of the game contradicts on the very next screen — and understates the odds, which is the difference between gambling looking like a guaranteed loss and looking like a fair bet.

## Suggested fix

Interpolate `DICE_WIN_MULTIPLIER` into Old Tom's answer the same way the gamble descriptor does, so the dialogue can't drift from the constant again.

### LINKED ISSUE #127: Tavern accepts a negative gambling bet, then contradicts itself with "You didn't bet any money!"
## Problem

`Tavern.changeBet` (`src/location/tavern.py:242-256`) only checks that the player can *afford* the amount:

```python
self.amount = int(amount)

if self.player.canAfford(self.amount):
    self.currentBet = self.amount
    self.currentPrompt.text = (
        "What will the dice land on? Current Bet: $%d" % self.currentBet
    )
```

`Player.canAfford` is `self.operatorMode or self.money >= cost` (`src/player/player.py:64-65`), so any negative amount passes — `money >= -100` is true for any balance. Entering `-100` sets `currentBet = -100` and the game reports **"What will the dice land on? Current Bet: $-100"**.

Picking a number then hits the guard in `Tavern.gamble` (`src/location/tavern.py:207`):

```python
if 1 <= input <= 6 and self.currentBet > 0:
```

which is false, so the player falls to the `else` branch and is told **"You didn't bet any money!"** — directly contradicting the bet the game just confirmed. There is no way to notice the bet was rejected other than trying to roll.

Not an exploit: because of the `self.currentBet > 0` guard the losing branch (`spendMoney(self.currentBet)`, which for a negative bet would *add* money) is unreachable. This is a validation and messaging bug, not a money bug.

## Suggested fix

Reject amounts below $1 in `changeBet` with a message that says so, alongside the existing "not enough money" branch, so the confirmed bet always matches what `gamble()` will accept.

## Test

`tests/location/test_tavern.py` already covers `changeBet` for the affordable and unaffordable cases; a negative-amount case belongs next to them.

dmccoystephenson:
## Summary
- `Tavern.changeBet` only checked affordability, so a negative bet amount passed `Player.canAfford` (`money >= negative` is always true) and set `currentBet` to a negative number. `gamble()`'s `currentBet > 0` guard then silently treated this as "no bet", directly contradicting the "Current Bet: $-100" prompt the game had just shown. Now rejects any amount below $1 with an explicit message.
- Old Tom's tavern dialogue still said gambling would "double your money", but the dice pay out `DICE_WIN_MULTIPLIER` (5x). Interpolated the constant so the dialogue can't drift from the payout again.
- Adds real-loop regression tests driving `Tavern.gamble()` itself (loss path, no-bet-placed path) and `Tavern.getDrunk()`'s money-loss branch, plus regression tests for both fixes above.

## Test plan
- [x] Full suite green: 380 passed (`SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest`)
- [x] Confirmed each new/changed test fails without its corresponding fix (stashed `src/location/tavern.py`, reran `tests/location/test_tavern.py`, saw the new bet-rejection and dialogue tests fail, then restored and confirmed pass)

Closes #127
Closes #126

---

### DIFF
--- a/src/location/tavern.py
+++ b/src/location/tavern.py
@@ -68,7 +68,7 @@ class Tavern:
                     "response": "Ah, the tavern! This is the place to unwind after a long day. "
                     "You can get yourself drunk for $10 - though you'll wake up at home with a headache the next day! "
                     "Or if you're feeling lucky, you can gamble with the dice. Place a bet, pick a number from 1 to 6, "
-                    "and if the dice matches your choice, you'll double your money!",
+                    "and if the dice matches your choice, you'll win %dx your money!" % DICE_WIN_MULTIPLIER,
                 },
                 {
                     "question": "Tell me about the other villagers.",
@@ -246,7 +246,11 @@ class Tavern:
             return
         self.amount = int(amount)
 
-        if self.player.canAfford(self.amount):
+        if self.amount < 1:
+            self.currentPrompt.text = (
+                "You have to bet at least $1! Money: $%.2f" % self.player.money
+            )
+        elif self.player.canAfford(self.amount):
             self.currentBet = self.amount
 
             self.currentPrompt.text = (
--- a/tests/location/test_tavern.py
+++ b/tests/location/test_tavern.py
@@ -31,6 +31,21 @@ def test_initialization():
     assert tavernInstance.npc.name == "Old Tom the Barkeep"
 
 
+def test_npc_tavern_dialogue_matches_dice_win_multiplier():
+    # prepare - regression test for #126: Old Tom's dialogue used to say
+    # gambling would "double your money", contradicting the actual payout.
+    from src.location.tavern import DICE_WIN_MULTIPLIER
+
+    tavernInstance = createTavern()
+
+    # call
+    response = tavernInstance.npc.get_dialogue_response(2)
+
+    # check
+    assert "double your money" not in response
+    assert "win %dx your money" % DICE_WIN_MULTIPLIER in response
+
+
 def test_run_get_drunk_action_success():
     # prepare
     tavernInstance = createTavern()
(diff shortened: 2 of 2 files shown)


dmccoystephenson:
Self-review rubric:
- Scope: PASS — only `src/location/tavern.py` and `tests/location/test_tavern.py` touched, both directly implementing #126/#127.
- Tests-new: PASS — bet-rejection (`test_changeBet_negative_amount_rejected`, `test_changeBet_zero_amount_rejected`) and dialogue (`test_npc_tavern_dialogue_matches_dice_win_multiplier`) each have a dedicated test; also added real-loop characterization tests for `gamble()`'s loss/no-bet paths and `getDrunk()`'s money-loss branch (these were pre-existing uncommitted WIP found in the tree at cycle start — see note below).
- Tests-fix: PASS — empirically verified: swapped `tavern.py` for the pre-fix version from `main`, reran `tests/location/test_tavern.py`, confirmed the 3 new regression tests (dialogue + both bet-rejection tests) FAIL with the expected assertion errors (`-100 == 0`, `'You have to bet at least $1' not in '...Current Bet: $0'`, `'double your money' is contained`), then restored the fix and confirmed all 29 tests in the file PASS.
- Sibling structure: PASS — new tests follow the `# prepare` / `# call` / `# check` convention and neighbor naming used throughout the file.
- Docs: PASS — no README/PLANNING references to the gambling payout wording or multiplier exist to drift.
- Issue resolution: PASS — #127's `changeBet` now rejects amounts below $1 before the affordability check; #126's dialogue now interpolates `DICE_WIN_MULTIPLIER` instead of a hardcoded "double".
- CI: PASS — green (https://github.com/Stephenson-Software/FishE/actions/runs/30247045534).
- Money-format: PASS — new rejection message uses `%.2f` for money, consistent with sibling messages in the same function.
- Deterministic-tests: PASS — all new tests patch `random` or avoid it entirely.

Note: at the start of this cycle the working tree had uncommitted, never-committed changes to `tests/location/test_tavern.py` (3 new tests) left over from an interrupted prior session. Two of them (`test_gamble_loss_via_real_loop`, `test_gamble_no_bet_placed`) had a latent bug — a mock `side_effect` that checked an accumulator list's truthiness as a "first call" flag without ever setting it on that branch, so the mocked `showOptions` returned the same input forever. Running them caused `Tavern.gamble()`'s `while True` loop to spin permanently, and the pytest process was observed consuming >19GB of RAM before being killed. Fixed by replacing the flawed sentinel with an explicit call counter in both tests before including them in this PR. Investigating and fixing this seemed safer than either silently discarding untested WIP or shipping it as-is.

---

PR #129, opened 2026-07-27
Show short game pauses through the front-end

### LINKED ISSUE #125: Fishing and getting drunk print flavour text to stdout, freezing the pygame and web front-ends
## Problem

Two pieces of game logic write flavour text straight to the terminal with `print()` and then block on `time.sleep()`, bypassing the user-interface contract entirely:

- `src/location/docks.py:338-340` — `print("Fishing... ")`, `sys.stdout.flush()`, `time.sleep(1)`
- `src/location/tavern.py:159-162` — a 3-iteration loop of `print("... ")`, `sys.stdout.flush()`, `time.sleep(1)`

On the console front-end this reads fine. On the other two front-ends it doesn't:

- **pygame:** the text goes to the terminal behind the game window, which the player isn't looking at. `Tavern.getDrunk` does call `self.userInterface.lotsOfSpace()` and `divider()` first (`src/location/tavern.py:157-158`), but `PygameUserInterface.lotsOfSpace` only fills the surface black and `divider()` is a no-op — neither calls `pygame.display.flip()`, so nothing is drawn either. The player sees the previous screen frozen for 3 seconds (1 second at the docks) with no explanation.
- **web:** `WebUserInterface.lotsOfSpace` and `divider` are documented no-ops, and nothing is presented, so the browser keeps rendering the previous screen for the same 1-3 seconds while "Fishing... " lands in the server's terminal.

This is exactly the front-end parity rule the codebase states in `src/ui/baseUserInterface.py`: everything player-visible should go through the front-end primitives so all three UIs stay in sync.

## Suggested fix

Add a `showBusy(message, seconds)` primitive to `BaseUserInterface` (with a concrete default so a future front-end still works without changing it) and implement it in all three UIs:

- `UserInterface` — print the message and the progress dots, as today.
- `PygameUserInterface` — draw the message and `flip()`, pumping events while it waits so the window stays responsive.
- `WebUserInterface` — present a `busy` screen so the browser shows the message; do not wait on the input queue.

Then have `Docks.fish` and `Tavern.getDrunk` call it instead of `print`/`sleep`.

## Impact

Cosmetic, but it is the only place where game logic writes player-facing text outside the UI contract, and it produces a "the game hung" experience on two of the three front-ends.

dmccoystephenson:
## Summary

Adds a **`showBusy(message, seconds)` primitive to `BaseUserInterface`** and routes the game's two short pauses through it.

`Docks.fish` and `Tavern.getDrunk` used to `print()` their flavour text and `time.sleep()` directly, bypassing the UI contract. On the console that read fine; on the other two front-ends the text went to the terminal and the player just saw the previous screen frozen for 1-3 seconds. (`getDrunk` also called `lotsOfSpace()`/`divider()` first, which on pygame fills the surface without ever flipping the display, and on web are documented no-ops — so nothing was drawn either.)

Implemented in all three front-ends:

- **console** — prints the message, then a dot row per second waited (the old "... " animation).
- **pygame** — draws the wrapped message and flips every frame, pumping events (including QUIT and resize) so the window stays responsive rather than looking hung.
- **web** — publishes a `busy` screen the client renders, and consumes no input; the next screen supersedes it.
- The base implementation is concrete (just the pause), so a future front-end still works without changing it — the abstract primitive list is unchanged, and `README.md`'s "implement `BaseUserInterface` + a `UIType` + a factory branch" instruction still holds as written.

Closes #125

> **Note:** this branch originally also carried fixes for #126 and #127, which landed independently in #128 while it was in review. It has been rebased onto that work, taking main's version of both; what remains is only the `showBusy` change.

## Test plan

- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest --verbose -vv --cov=src --cov-report=term-missing --cov-report=xml:cov.xml` — **388 passed**, coverage 95%
- [x] **All three front-ends touched and covered:** new tests for the console (`test_userInterface.py`), pygame (`test_pygameUserInterface.py` — the message is drawn, a keypress neither ends the pause nor rewrites the prompt, and an over-wide message wraps), web (`test_webUserInterface.py` — the `busy` screen is presented, no input is consumed, and the client page has a matching render branch), and the inherited default (`test_baseUserInterface.py`).
- [x] Call-site tests updated: `test_docks.py`/`test_tavern.py` stub `showBusy` instead of patching `print`/`sys.stdout.flush`/`time.sleep`, and assert the pause is announced through the UI.
- [x] No `Player`/`Stats`/`TimeService` field changed, so `schemas/*.json` and the `*JsonReaderWriter`s are correctly untouched; `README.md`/`PLANNING.md` describe nothing that changed here.

---

### DIFF
--- a/src/location/docks.py
+++ b/src/location/docks.py
@@ -1,6 +1,4 @@
 import random
-import sys
-import time
 
 from location.enum.locationType import LocationType
 from player.player import Player
@@ -332,12 +330,7 @@ class Docks:
         return 1.0, ""
 
     def fish(self):
-        self.userInterface.lotsOfSpace()
-        self.userInterface.divider()
-
-        print("Fishing... "),
-        sys.stdout.flush()
-        time.sleep(1)
+        self.userInterface.showBusy("Fishing...", 1)
 
         # Capture the time of day at the start of the trip (the loop advances it).
         timeFactor, timeLabel = self.getTimeOfDayModifier(self.timeService.time)
--- a/src/location/tavern.py
+++ b/src/location/tavern.py
@@ -1,6 +1,4 @@
 import random
-import sys
-import time
 
 from location.enum.locationType import LocationType
 
@@ -68,7 +66,8 @@ class Tavern:
                     "response": "Ah, the tavern! This is the place to unwind after a long day. "
                     "You can get yourself drunk for $10 - though you'll wake up at home with a headache the next day! "
                     "Or if you're feeling lucky, you can gamble with the dice. Place a bet, pick a number from 1 to 6, "
-                    "and if the dice matches your choice, you'll win %dx your money!" % DICE_WIN_MULTIPLIER,
+                    "and if the dice matches your choice, you'll win %dx your money!"
+                    % DICE_WIN_MULTIPLIER,
                 },
                 {
                     "question": "Tell me about the other villagers.",
@@ -151,15 +150,9 @@ class Tavern:
             return LocationType.DOCKS
 
     def getDrunk(self):
-        self.userInterface.lotsOfSpace()
-        self.userInterface.divider()
-
         self.player.spendMoney(10)
 
-        for i in range(3):
-            print("... ")
-            sys.stdout.flush()
-            time.sleep(1)
+        self.userInterface.showBusy("You drink your way into the evening...", 3)
 
         self.stats.timesGottenDrunk += 1
 
--- a/src/ui/baseUserInterface.py
+++ b/src/ui/baseUserInterface.py
@@ -1,3 +1,4 @@
+import time
 from abc import ABC, abstractmethod
 from prompt.prompt import Prompt
 from player.player import Player
@@ -88,6 +89,16 @@ class BaseUserInterface(ABC):
     def cleanup(self):
         """Release any resources held by the front-end."""
         pass
+
+    def showBusy(self, message, seconds=1.0):
+        """Show a message while the game pauses, without waiting for input.
+
+        Used by the short "Fishing..." style beats, which otherwise have no way
+        to say anything: showDialogue would demand a keypress the game never
+        asked for. Concrete (not abstract) so a front-end that has nothing to
+        draw still gets the pause; every front-end in this repo overrides it to
+        actually show the message."""
+        time.sleep(seconds)
 
     def promptForNumber(self, promptText):
         """Prompt for a number via promptForText; return a float or None if the
--- a/src/ui/pygameUserInterface.py
+++ b/src/ui/pygameUserInterface.py
@@ -478,6 +478,37 @@ class PygameUserInterface(BaseUserInterface):
 
         return entered
 
+    def showBusy(self, message, seconds=1.0):
+        """Draw a message and hold it for `seconds`, taking no input.
+
+        The window keeps being redrawn and its events pumped throughout, so the
+        pause looks like the game working rather than the window hanging."""
+        endTime = time.time() + seconds
+        while time.time() < endTime:
+            for event in pygame.event.get():
+                if event.type == pygame.QUIT:
+                    self.cleanup()
+                    sys.exit()
+                elif event.type == pygame.VIDEORESIZE:
+                    self._handle_resize(event.w, event.h)
+
+            self._draw_busy(message)
+            pygame.display.flip()
+            pygame.time.Clock().tick(60)
+
+    def _draw_busy(self, message):
+        """Draw the wrapped busy message. Split out from showBusy so the layout
+        can be tested without driving the wait loop."""
+        self.screen.fill(self.BLACK)
+        margin_x = self.width * 0.06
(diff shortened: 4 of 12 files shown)


dmccoystephenson:
## Self-review

Reviewed the full diff against this repo's conventions — front-end parity, the schema/reader-writer contract, test coverage on new public methods, and doc drift. **No blocking findings.** One weak test was found and fixed in this branch before posting (5f8cb4a). Notes below, in `path:line — finding` form.

### Fixed during review

`tests/ui/test_pygameUserInterface.py:494 — test_showBusy_takes_no_input_and_leaves_the_prompt_alone` stubbed `time.time` with `[0.0, 5.0]`, so the wait loop exited on its first condition check and `pygame.event.get()` was never reached. The injected keypress was never read, meaning the test proved only that the prompt was untouched — not the "takes no input" half of its name. Rewritten as `test_showBusy_ignores_keypresses_and_leaves_the_prompt_alone`, running two frames with the keypress in the first and asserting a second frame was still drawn afterward.

### Verified, no change needed

`src/ui/baseUserInterface.py:93 — showBusy is concrete, not abstract.` Deliberate: `tests/ui/test_baseUserInterface.py`'s `RecordingUserInterface` and any external front-end implement only the seven abstract primitives, so making this an eighth would break them. The abstract list is unchanged and `README.md`'s "implement `BaseUserInterface` + a `UIType` + a factory branch" instruction still holds as written. All three in-repo front-ends override it anyway.

`src/ui/webUserInterface.py:339 — showBusy leaves anything already in the input queue for the next screen.` Considered and accepted: the client can't send during a busy screen — there is no button, and the `keydown` handler only acts on `options`/`dialogue`/`timed`, so a `busy` screen falls through it. `tests/ui/test_webUserInterface.py` pins that behaviour rather than leaving it implicit.

`src/ui/webUserInterface.py:160 — the busy screen carries no header`, so the status chips disappear during the pause. That matches the existing `dialogue`, `prompt`, and `timed` screens, which also omit `header`, and the console/pygame front-ends likewise drop the status block on non-menu screens. Consistent, so left alone.

`src/location/tavern.py:152 — getDrunk now spends the $10 before the pause rather than after clearing the screen.` No behavioural difference: the charge still lands before `increaseDay()` rolls the day over, and `run()` still gates the call on `canAfford(10)`.

`src/location/docks.py:333 — fish() can still return early ("You're too tired to fish!") after the pause has been shown.` Unchanged from before this PR, where the `print`/`sleep` ran in exactly the same position.

`src/location/tavern.py:242 — the < 1 check runs after int(amount)`, so `0.5` and `-0.5` both truncate to `0` and are rejected. Covered by `test_changeBet_rejects_a_bet_rounding_down_to_zero`, with `test_changeBet_accepts_the_smallest_valid_bet` pinning that `$1` still goes through.

`src/ui/userInterface.py:133 — console dot placement changed slightly` (message first, then one dot row per second, where the docks previously printed `"Fishing... "` with the dots inline and no trailing row). Cosmetic, and it makes the two call sites consistent with each other.

### Contract checks

- **Front-end parity:** implemented in all three UIs plus a base default; each has its own test.
- **Schema / reader-writer:** no `Player`/`Stats`/`TimeService` field added, renamed, or retyped, so `schemas/*.json` and the `*JsonReaderWriter`s are correctly untouched.
- **Docs:** `README.md` and `PLANNING.md` describe nothing that changed here (neither mentions the dice payout or the pauses); re-checked both.
- **Build:** `python3 -m compileall -q src tests` clean; `pytest --cov` 386 passed, 94%.

dmccoystephenson:
### Rebase note (scope reduced)

#128 landed on `main` while this was in review and fixed #126 and #127 independently. This branch has been rebased onto it, taking **main's version of both** — its dialogue wording (`"you'll win 5x your money!"`) and its bare `< 1` guard without my explanatory comment — and dropping my duplicate regression tests in favour of the ones #128 already added (`test_npc_tavern_dialogue_matches_dice_win_multiplier`, `test_changeBet_negative_amount_rejected`, `test_changeBet_zero_amount_rejected`). What remains here is only the `showBusy` work for #125.

Two incidental changes came out of the rebase, both worth calling out:

- `src/location/tavern.py:69 —` #128's dialogue line was 97 characters, over black's limit; re-wrapped by the repo's own `format.sh`.
- `tests/location/test_tavern.py:526,555 —` same, for two `MagicMock(side_effect=showOptionsSideEffect)` lines in #128's new tests.
- `tests/location/test_tavern.py:571 — test_getDrunk_loses_money_in_drunken_stupor` (added by #128) stubbed `tavern.print`, `tavern.sys.stdout.flush`, and `tavern.time.sleep`. `tavern.py` no longer imports `sys` or `time`, so those stubs would now raise `AttributeError`; they're removed, and the shared `createTavern()` helper stubs `showBusy` instead. The test's own assertions are unchanged and still pass.

Suite re-run after the rebase: **388 passed**, coverage 95%.

PR #136, opened 2026-08-01
Hire villagers as crew, unlocking new dialogue

dmccoystephenson:
## Summary

Crew slots used to be an anonymous headcount — "Hire a Worker" just incremented `player.workers`. Hiring is now a choice of *which villager* joins the outfit, and every hire unlocks conversations the player couldn't have before.

- **A named workforce.** `src/npc/villagers.py` adds a 12-villager roster, each with a backstory, a specialty, and a one-line blurb shown next to the wage in the hire menu. The roster is deliberately at least as long as the largest boat's crew capacity, so every berth on a Fishing Fleet can be filled by name.
- **Conditional dialogue.** `NPC` dialogue options now accept an optional `"condition"` — a zero-arg callable — and stay hidden while it returns `False`. `get_dialogue_response` indexes the *available* options, i.e. the same list the front-ends number their menus from, so a hidden option can never shift a response onto the wrong question.
- **Talk to Your Crew.** A new docks option appears as soon as anyone is aboard. Each crew member answers about their work, the boat, and whether wages are turning up on time, with three further questions that unlock on a full crew, a named business, and a Fishing Fleet respectively.
- **The villagers you already know react.** Sam sizes up who you hired (naming them and their specialty), Gilbert notices them spending wages across his counter, and Old Tom relays what they say at the bar — including a warning when payroll is short. All three are gated behind having hired anyone.
- **Persistence.** `player.hiredWorkers` round-trips through `PlayerJsonReaderWriter` and `schemas/player.json`. `workers` stays the authoritative headcount the production maths runs on, so a save written before crews had names loads with its crew intact — those hands are just unnamed, and are dismissed (or quit over unpaid wages) before any named villager is.
- `business.hireWorker` / `dismissWorker` centralise the roster/headcount invariant; `sellBoat` and the unpaid-wages quit path keep the two in sync, and `runDailyProduction`'s summary now reports `quitNames`.

## Test plan

- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest --verbose -vv --cov=src --cov-report=term-missing --cov-report=xml:cov.xml` — 463 passed. New/changed lines are fully covered (`src/npc/villagers.py` 100%, `src/npc/npc.py` 100%, `src/player/*` 100%, `src/location/docks.py` 98% with only pre-existing `fish()` branches uncovered).
- [x] `black` + `autoflake` run over the changed files only (unrelated files the formatter also touches were left alone to keep the diff readable).
- [x] **Front-end parity.** Everything new goes through the shared primitives — `showOptions`, `showDialogue`, `showInteractiveDialogue` — so all three front-ends are covered: pygame and web inherit `BaseUserInterface.showInteractiveDialogue`, and the console overrides it. Both paths re-read `get_dialogue_options()` each turn, and each has its own test that a conditional option is hidden until unlocked (`tests/ui/test_baseUserInterface.py`, `tests/ui/test_userInterface.py`).
- [x] Played the hire → talk → dismiss flow manually through the real console front-end.

## Docs

`README.md`'s Fishing Business section was rewritten to describe hiring by name and the dialogue it unlocks. `PLANNING.md` and the other schemas were checked and needed no change.

---


src/npc/npc.py
```diff
@@ -11,21 +11,41 @@ def __init__(self, name: str, backstory: str, dialogue_options: list = None):
...
+        from - so conditional options never shift a response onto the wrong
+        question.
+
         A response may be a plain string, or a zero-arg callable that's
         evaluated on demand - letting a response reflect current game state
         (e.g. an NPC commenting on the player's fishing business) instead of
         being fixed at NPC-construction time."""
-        if 0 <= option_index < len(self.dialogue_options):
-            response = self.dialogue_options[option_index].get("response", "")
+        options = self.get_dialogue_options()
```
dmccoystephenson, line 39: This is the one behavioural change to an existing contract: `get_dialogue_response` used to index `self.dialogue_options` directly and now indexes the filtered list. That's required for correctness — every front-end numbers its menu from `get_dialogue_options()`, so with a hidden option present the old indexing would return the wrong answer. For any NPC without conditions the two lists are identical, so existing callers are unaffected.

src/location/docks.py
```diff
@@ -141,6 +154,10 @@ def run(self):
             self.manageBusiness()
             return LocationType.DOCKS
 
+        elif input == "8" and self.player.hiredWorkers:
```
dmccoystephenson, line 157: The `and self.player.hiredWorkers` guard is redundant today — all three front-ends validate the choice against the option list they were handed (console loops on a mismatch, web rejects anything outside `valid`, pygame bounds-checks the number keys and wraps the arrow selection), so `"8"` can only come back when the option was actually offered. Kept anyway so the branch can't fire against an empty crew if a future front-end is looser about it.

src/location/docks.py
```diff
@@ -97,6 +105,11 @@ def run(self):
             "Go to Bank",
             "Manage Boat & Crew",
         ]
+        # Appended (rather than slotted in next to the other "Talk to" option)
+        # so the numbering of everything above stays put whether or not the
+        # player has anyone to talk to.
+        if self.player.hiredWorkers:
```
dmccoystephenson, line 111: Worth knowing: this gates on `hiredWorkers`, not `workers`. A save from before this change loads with a crew but no names, so those players see no "Talk to Your Crew" until they hire someone new — at which point the option appears and Sam's line mentions the unnamed hands alongside the named one. That's the intended read of an old save rather than an oversight; the alternative (backfilling names onto existing headcount) would invent history the save never had.

src/business/business.py
```diff
@@ -60,6 +63,50 @@ def sellBoat(player):
     player.hasBoat = False
     player.boatTier = 0
     player.workers = 0
+    player.hiredWorkers = []
+    return True
+
+
+def _trimCrewRoster(player):
+    """Drop named crew until the roster fits the headcount, and return the
```
dmccoystephenson, line 71: `_trimCrewRoster` is the single place the roster/headcount invariant (`len(hiredWorkers) <= workers`) is enforced, called from `dismissWorker` and the unpaid-wages path in `runDailyProduction`. Popping from the tail gives the desired ordering for free: unnamed legacy hands are absorbed by the headcount drop before any named villager is asked to leave, and among named ones the most recent hire goes first.

dmccoystephenson (commented):
Self-review (no configured reviewer on this repo). Read the full diff back; the suite is green at 463 tests with the new modules fully covered. Three notes below on deliberate choices a reader might otherwise flag, plus one behaviour worth knowing about. Nothing here blocks merge.

PR #138, opened 2026-08-01
Export fish to other villages with a Trawler or better

dmccoystephenson:
## Summary

Gilbert's shop only spends `SHOP_DAILY_BUDGET` ($750) a day. A maxed-out Fishing Fleet crew lands ~120 fish a day, worth roughly $760 at average species values — so a full crew saturates the village's only buyer exactly, and everything the player catches themselves on top of that just piles up unsellable. This adds a second sales channel gated behind the boat ladder.

**Three markets**, each trading a bigger premium for a bigger freight bill:

| Market | Boat | Price | Freight |
|---|---|---|---|
| Saltmarsh | Trawler | ×1.2 | $25 |
| Kestrel Cove | Trawler | ×1.5 | $250 |
| Thornhaven | Fishing Fleet | ×2.0 | $900 |

Because freight is flat and the premium is proportional, the best market depends on the size of the load — the break-evens land at roughly 120 fish (Saltmarsh → Kestrel Cove) and 205 fish (Kestrel Cove → Thornhaven), so all three have a real niche rather than one dominating. The menu shows what the current hold would fetch at each before the player commits.

**What limits a run** (since the markets themselves have no daily budget):
- **The hold.** `exportCapacity` is new on `BOAT_TIERS`: a Rowboat is 0 and can't cross at all, a Trawler carries 250 per run, a Fishing Fleet 600. The best fish load first; the rest wait for the next run.
- **The freight**, charged *up front* rather than netted out of the proceeds — `money` has a schema minimum of 0, so deducting from proceeds could produce an unsaveable player. A run the player can't fund is refused with a message naming the cost, the shortfall, and what to do instead.
- **A day.** The round trip calls `increaseDay`, so exporting isn't a free repeatable action — the crew fish, wages come due and rent falls while the player is away (eviction is reported in the trip summary, same as after a night at the tavern).

Per-day, this works out at roughly 1.7× the shop's ceiling for a Fishing Fleet owner once accumulation time is accounted for — a real answer to the bottleneck without trivialising the $10,000 goal, which a $6,000 boat already puts the player most of the way toward.

**Supporting changes**
- Extracted `fish.bestFirst` and `Player.removeFish`, now shared by the shop's budget-limited sale and the export hold's capacity-limited one (the two had duplicated the same sort-and-decrement logic).
- Rebuilt the docks menu as parallel options/actions lists — the same pattern `manageBusiness` already uses in that file. With two conditional entries ("Talk to Your Crew" and the new export option), dispatching on hardcoded numbers would eventually fire the wrong branch; there's a regression test for exactly that case.
- New lifetime stats (`totalFishExported`, `totalMoneyFromExports`, `totalShippingPaid`) with matching `schemas/stats.json` fields and two milestones (First Export, Coastal Trader).
- Gilbert gets a question — unlocked once the player's boat can actually reach the markets — explaining his daily budget and pointing outward, and the "shop is out of money" message now suggests shipping the rest rather than only waiting a day.

## Test plan

- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest --verbose -vv --cov=src --cov-report=term-missing --cov-report=xml:cov.xml` — 514 passed. `src/business/export.py`, `src/fish/fish.py`, `src/player/player.py` and both stats modules are at 100%; `docks.py`/`shop.py` at 99%, with only pre-existing `fish()` branches uncovered.
- [x] `black` + `autoflake` over the changed files only.
- [x] **Front-end parity.** Everything new goes through `showOptions` and `currentPrompt`, which all three front-ends already implement — no new primitive, no front-end-specific path. Legacy saves are covered too: an untyped hold (no `fishByType`) ships and is priced at the original flat range.
- [x] Played the flow through the real console front-end: a full Fishing Fleet run, a small load reading as a loss, an unaffordable-freight refusal, and an empty hold.

## Docs

`README.md` gains an "Exporting to Other Villages" section; the Selling Fish and Milestones sections were updated to match. `PLANNING.md` needed no change.

---


src/location/docks.py
```diff
@@ -119,45 +135,52 @@ def run(self):
         descriptor += " " + self._weatherDescriptor()
         input = self.userInterface.showOptions(descriptor, li)
 
-        if input == "1":
+        choice = int(input)
+        action = actions[choice - 1]
```
dmccoystephenson, line 139: This is the riskiest change in the PR — the docks menu no longer dispatches on hardcoded `input == "7"` strings. It had to change: with two independently-conditional entries, `"8"` means the crew when only the crew entry is present and the export menu when only that one is, and the old `elif input == "8" and self.player.hiredWorkers` would silently do nothing in the second case. `test_run_menu_positions_hold_when_both_extras_are_present` and `test_run_crew_action_still_fires_without_the_export_entry` pin both arrangements. The fixed seven keep their original numbers, which is why every pre-existing docks test still passes untouched.

src/business/export.py
```diff
@@ -0,0 +1,148 @@
...
+        return summary
+    if player.fishCount <= 0:
+        summary["reason"] = "empty_hold"
+        return summary
+    if not player.canAfford(market["shippingCost"]):
+        summary["reason"] = "cannot_afford_freight"
+        return summary
+
+    cargo = buildCargo(player)
+    player.spendMoney(market["shippingCost"])
```
dmccoystephenson, line 129: Freight is deducted here, before the sale, rather than being netted out of the proceeds — deliberate, and not just stylistic. `schemas/player.json` requires `money >= 0`, so a losing run that subtracted the fee from the payout could drive the balance negative and produce a player that fails schema validation on the next save. Charging up front means the `canAfford` guard above is the only place that can block a run, and `test_runExport_never_puts_the_player_into_debt` covers the worst case the menu allows (one Minnow against Thornhaven's $900).

src/business/export.py
```diff
@@ -0,0 +1,148 @@
...
+    the midpoint of that range - it's shown to the player before they commit,
+    where an honest estimate matters more than a precise one."""
+    gross = 0.0
+    for species in cargo:
+        fishType = fish.getFishType(species)
+        if fishType is None:
+            # Legacy untyped fish, priced at the middle of the old $3-5 range.
+            midpoint = 4.0
+        else:
+            midpoint = (fishType["minValue"] + fishType["maxValue"]) / 2
```
dmccoystephenson, line 92: The estimate shown in the menu uses each species' midpoint while the actual sale rolls `fishValue` per fish, so the two differ by a few percent on a real load (measured ~$6280 estimated vs ~$6230 actual on a 600-fish hold). That's intended — the alternative is either rolling the prices early and holding them, or showing a range — but it's why the option text says "about $X clear" rather than quoting a figure the player could hold us to.

src/location/shop.py
```diff
@@ -178,20 +201,10 @@ def sellFish(self):
...
-            for species, count in self.player.fishByType.items():
-                queue.extend([species] * count)
-            queue.sort(
-                key=lambda s: (fish.getFishType(s) or {}).get("maxValue", 0),
-                reverse=True,
-            )
-        else:
-            # Legacy save with only an aggregate count (no species breakdown).
-            queue = [None] * self.player.fishCount
+        queue = fish.bestFirst(self.player.fishByType, self.player.fishCount)
```
dmccoystephenson, line 207: This replaces ten lines of inline queue-building that were byte-for-byte the logic export needs, so it moved to `fish.bestFirst`. Worth flagging because it touches the existing sale path rather than only adding to it: the behaviour is identical (same sort key, same legacy `[None] * fishCount` fallback), and the shop's own budget tests plus the new `tests/fish/test_fish.py` cases cover it from both sides.

dmccoystephenson (commented):
Self-review (no configured reviewer on this repo). Read the full diff back; 514 tests green, new modules at 100% coverage. Four notes on decisions a reader would reasonably question, none blocking.

PR #148, opened 2026-08-01
Run the web game in the player's browser, with saves in IndexedDB

dmccoystephenson:
## Summary

FishE's web front-end ran the game **on the server** and wrote save slots to the server's disk, so every visitor to a deployment shared one game loop and one set of save files. This adds a second browser front-end that runs the whole Python game **in the player's own tab** under Pyodide, with that tab's save slots kept in the browser's IndexedDB — the same approach Roam uses for its web build.

The server-backed front-end (`UIType.WEB`, `examples/web_app.py`) is unchanged and still there; the deployed image now serves the browser-native one.

### Front-end parity

`PyodideUserInterface` subclasses `WebUserInterface` and overrides only the two transport seams introduced here, `_present` and `_awaitInput`. Every screen — options, dialogue, prompt, busy, timed, ended — is still defined exactly once, so the two browser front-ends cannot drift.

The browser half is shared the same way: the renderer and styles moved out of `HTML_PAGE` into `web/client.js` and `web/client.css`, which the server-backed page inlines and the Pyodide page links. `tests/web/test_clientParity.py` asserts the sharing itself.

### Why a SharedArrayBuffer

FishE's game loop is synchronous, so it blocks the Worker whenever it waits for the player — and a blocked Worker can never run an `onmessage` handler. Shared memory needs no event loop, so that is how input reaches a blocked game. This is also why `web/serve.py` must send the COOP/COEP headers: without cross-origin isolation, `SharedArrayBuffer` does not exist and nothing the player clicks reaches the game. `web/index.html` says exactly that on screen if it happens.

For the same reason, IndexedDB writes happen on the main thread rather than in the Worker: the Worker posts the save file map over, and the page stores it.

### Saves

- `browserSaveSync.syncBrowserSaves()` flushes the save directory through the Worker's `syncSaves` hook after every save write, slot deletion and legacy migration. It is a no-op outside Pyodide, where the files on disk are the real saves.
- `FISHE_SAVE_DIR` relocates the save directory (the Worker points it at its IndexedDB-backed `/saves`).
- Deleting a slot in-game clears and rewrites the store, so a deletion sticks across a reload.

### Deployment

`web/serve.py` serves the bundle and never runs the game. The Dockerfile serves that instead of `examples/web_app.py`, which means the image needs no `pip install` (the game's one dependency, `jsonschema`, is loaded browser-side) and no save volume at all.

## Test plan

- [x] `python3 -m compileall -q src tests web`
- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest --verbose -vv --cov=src --cov-report=term-missing --cov-report=xml:cov.xml` — **716 passed**, 97% coverage (`src/ui/pyodideUserInterface.py` 97%)
- [x] `python3 -m black --check src tests web` clean; `autoflake` run per `format.sh`
- [x] New tests: `tests/ui/test_pyodideUserInterface.py` (18), `tests/web/` (13), `tests/test_browserSaveSync.py` (4), plus save-flush and `FISHE_SAVE_DIR` cases
  - Includes an end-to-end test that boots a **real** `FishE` through the Pyodide front-end against a fake `js` module whose `syncSaves` mirrors `web/game-worker.js`, and asserts the saved game lands in simulated browser storage
  - Includes a ring-buffer wraparound test and a test that quitting ends the page instead of surfacing `SystemExit` as a Worker error
- [x] Both browser front-ends touched and verified; console and pygame untouched
- [x] All four JS blocks (`client.js`, `game-worker.js`, and both inline transport scripts) pass `node --check`
- [x] `web/serve.py` smoke-tested live: `/` and `/play` serve the page with COOP/COEP set, `/web/*` serves the client, worker and 98 KB bundle, and `/src/fishE.py` is refused with a 404
- [ ] Not verified: an actual browser loading Pyodide and playing a save through IndexedDB — there is no browser in the environment this was built in. The Python side of that path is covered by the end-to-end test above, and the JavaScript side mirrors Roam's proven implementation, but the first real load is worth watching.

No schema or save-format change: `Player`, `Stats` and `TimeService` are untouched, so `schemas/*.json` and the `*JsonReaderWriter` classes stay in sync.

---

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,37 +2,38 @@ FROM python:3.12-slim
 
 WORKDIR /app
 
-# jsonschema is a real runtime dependency, not a test-only one: src/fishE.py
-# imports it at module scope, so the game cannot start without it. pygame is
-# deliberately absent — it is imported lazily by the pygame front-end only,
-# which this image never selects.
-COPY requirements.txt ./
-RUN pip install --no-cache-dir -r requirements.txt
-
+# This image does not run the game — the browser does. web/serve.py only hands
+# out the page, the Worker, and the game bundle; Pyodide then runs the Python
+# game in the player's own tab, with that tab's IndexedDB holding the saves.
+#
+# Two consequences worth spelling out:
+#   - No pip install. The game's one runtime dependency, jsonschema, is loaded
+#     browser-side by web/game-worker.js, and the server here is pure stdlib.
+#   - No save directory, and no volume to persist. Every visitor gets their own
+#     game and their own save slots, which is exactly what the server-backed
+#     front-end (examples/web_app.py) could not give them.
 COPY src/ ./src/
-COPY examples/ ./examples/
+COPY web/ ./web/
 # The JSON Schemas the save-file readers validate against. Their paths are
-# relative to the process cwd (see playerJsonReaderWriter.PLAYER_SCHEMA_PATH =
-# "schemas/player.json"), which is /app here — so they must be copied in, or
-# every save load/write raises FileNotFoundError.
+# relative to the process cwd, which the Worker sets to the unpacked bundle —
+# so these have to be copied in for build_zip.py to put them in the bundle, not
+# for anything in this image to read.
 COPY schemas/ ./schemas/
+COPY version.txt ./
 
-# Create the data/ directory (see saveFileManager.py's default
-# data_directory="data", resolved relative to the process cwd -> /app/data
-# here) and hand it to a non-root user before switching to it, so a named
-# volume mounted over /app/data on first run inherits writable ownership.
-RUN useradd --system --no-create-home fishe \
-    && mkdir -p /app/data \
-    && chown -R fishe:fishe /app/data
+# Bundle the game for the browser to download. Built here rather than checked
+# in so the bundle can never be a stale copy of src/.
+RUN python3 web/build_zip.py
+
+RUN useradd --system --no-create-home fishe
 USER fishe
 
-# WebUserInterface itself defaults to 127.0.0.1:8000 (unreachable from
-# outside its own network namespace); UserInterfaceFactory's WEB branch reads
-# these to override the bind address/port. 0.0.0.0 is required for Traefik
-# (a different container on the same bridge network) to reach this server.
+# web/serve.py defaults to 127.0.0.1 (unreachable from outside its own network
+# namespace), so 0.0.0.0 is required for Traefik — a different container on the
+# same bridge network — to reach it.
 ENV FISHE_WEB_HOST=0.0.0.0
-ENV FISHE_WEB_PORT=8000
+ENV FISHE_WEB_PORT=8080
 ENV PYTHONUNBUFFERED=1
 
-EXPOSE 8000
-CMD ["python3", "examples/web_app.py"]
+EXPOSE 8080
+CMD ["python3", "web/serve.py"]
--- a/src/fishE.py
+++ b/src/fishE.py
@@ -13,6 +13,7 @@ from stats.stats import Stats
 from ui.userInterfaceFactory import UserInterfaceFactory
 from ui.enum.uiType import UIType
 from saveFileManager import SaveFileManager
+from browserSaveSync import syncBrowserSaves
 from achievements import achievements
 from achievements.achievements import GOAL_AMOUNT, GOAL_MILESTONE_NAME
 from housing import housing
@@ -300,6 +301,12 @@ class FishE:
         except (IOError, OSError) as e:
             print(f"\n Warning: Failed to save game: {e}")
             # Game continues even if save fails
+            return
+
+        # Under the Pyodide front-end the writes above landed in the Worker's
+        # in-memory filesystem; this is what actually gets them into the
+        # browser's IndexedDB. A no-op for every other front-end.
+        syncBrowserSaves()
 
     def loadPlayer(self):
         try:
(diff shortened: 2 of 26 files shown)


dmccoystephenson (commented):
Self-review. Read the full diff; no correctness defects found, and CI is green (716 tests, 97% coverage). Four notes below record decisions a future reader is likely to second-guess, plus the one thing this change genuinely cannot verify from the environment it was built in.

The one open risk is stated in the PR description and repeated here: no browser was available to load the page, so Pyodide booting, `loadPackage(['jsonschema'])`, and a real IndexedDB round-trip are unverified end-to-end. Everything on the Python side of that path is covered by a test that runs a real `FishE` through `PyodideUserInterface` against a fake `js` module mirroring `game-worker.js`, and the JavaScript mirrors Roam's production implementation — but the first real page load should be watched rather than assumed.

One dependency of the design worth recording since it is invisible from the code: COEP `require-corp` blocks cross-origin subresources that do not opt in, and the Worker loads Pyodide from jsdelivr via `importScripts`. That was checked against the live CDN, which sends `cross-origin-resource-policy: cross-origin` — so the boot is not blocked by the very headers `SharedArrayBuffer` requires. If Pyodide is ever moved to another CDN, that property has to be rechecked.

src/fishE.py
```diff
@@ -300,6 +301,12 @@ def save(self):
         except (IOError, OSError) as e:
             print(f"\n Warning: Failed to save game: {e}")
             # Game continues even if save fails
+            return
```
dmccoystephenson, line 304: This early `return` is load-bearing rather than tidying. The three save files are written in sequence, so a failure partway through leaves the save directory in a mixed state; returning here skips the flush below, which leaves the previous, consistent set in browser storage rather than overwriting it with a half-written one. Worth keeping this ordering in mind if a fourth save file is ever added.

web/index.html
```diff
@@ -0,0 +1,155 @@
...
+      req.onerror   = (ev) => reject(ev.target.error);
+    });
+  }
+
+  // The file map is the whole save directory, so the store is cleared first —
+  // that is what makes deleting a save slot in-game actually stick.
+  function idbWrite(files) {
+    idbOpen().then((db) => {
+      let tx;
+      try { tx = db.transaction(IDB_STORE, "readwrite"); }
```
dmccoystephenson, line 112: The `clear()` and the `put()`s share one transaction deliberately: the file map is the whole save directory, so a delete only sticks if the store is emptied first — and doing both in one transaction means a tab closed mid-write aborts the transaction and keeps the previous saves, instead of leaving the player with a cleared store.

web/index.html
```diff
@@ -0,0 +1,155 @@
...
+
+  const sab     = new SharedArrayBuffer(8 + RING_SIZE);
+  const sabMeta = new Int32Array(sab, 0, 2);
+  const sabData = new Uint8Array(sab, 8, RING_SIZE);
+
+  function writeToRing(text) {
+    const bytes = new TextEncoder().encode(text + "\n");
+    const writeIndex = Atomics.load(sabMeta, 0);
+    const readIndex  = Atomics.load(sabMeta, 1);
+    if (bytes.length > RING_SIZE - (writeIndex - readIndex)) {
```
dmccoystephenson, line 67: This fullness check relies on the ring indices being monotonic, which makes it wrong after `writeIndex` overflows `Int32Array`'s range — about 2 GB of cumulative player input in a single page load. That is unreachable for a turn-based game where each message is a few dozen bytes and the game drains one per screen, so it is left as-is rather than complicated with wraparound arithmetic. Flagging it in case this ring is ever reused for something higher-volume, like mouse movement.

web/serve.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class _Server(http.server.ThreadingHTTPServer):
+    allow_reuse_address = True
+    daemon_threads = True
+
+
+def main():
+    host = os.environ.get("FISHE_WEB_HOST", "127.0.0.1")
+    port = int(os.environ.get("FISHE_WEB_PORT", "8080"))
```
dmccoystephenson, line 87: `FISHE_WEB_HOST`/`FISHE_WEB_PORT` are reused from the server-backed front-end on purpose — they are the same question, "where does FishE listen" — but the default differs (8080 here, 8000 there). Anything that pinned `FISHE_WEB_PORT=8000` for the old service will keep binding 8000 after switching to this entry point, which is silent rather than broken. The deployment that switches over has to move its port mapping and healthcheck together.

Dockerfile
```diff
@@ -2,37 +2,38 @@ FROM python:3.12-slim
...
+#
+# Two consequences worth spelling out:
+#   - No pip install. The game's one runtime dependency, jsonschema, is loaded
+#     browser-side by web/game-worker.js, and the server here is pure stdlib.
+#   - No save directory, and no volume to persist. Every visitor gets their own
+#     game and their own save slots, which is exactly what the server-backed
+#     front-end (examples/web_app.py) could not give them.
 COPY src/ ./src/
-COPY examples/ ./examples/
+COPY web/ ./web/
```
dmccoystephenson, line 16: Note for anyone reading this later: dropping `pip install` and `examples/` means this image can no longer run the server-backed front-end at all — overriding `CMD` back to `examples/web_app.py` fails on the missing `jsonschema`. That is the intent, not an oversight. The absence of a save volume is the point of the change: there is no server-side save state left to persist.

Issue #154, opened 2026-08-02
test: two tavern gamble tests re-implement the payout instead of calling gamble(), and encode the wrong multiplier


dmccoystephenson:
## Summary

During a triage pass, two tests in `tests/location/test_tavern.py` were found to perform the arithmetic they then assert on, without ever invoking the code under test.

`tests/location/test_tavern.py:248-295` (`test_gamble_win_shows_correct_amount`) sets up a mock, abandons it, and hand-executes the win branch:

```python
# We need to capture the state after the win but before the next iteration
# Let's test the win logic directly instead
input_value = 1
tavernInstance.diceThrow = 1

# Execute the win condition logic
winAmount = tavernInstance.currentBet
tavernInstance.player.money += tavernInstance.currentBet
tavernInstance.stats.totalMoneyMade += tavernInstance.currentBet
...
assert tavernInstance.player.money == 150  # Won 50
```

`tests/location/test_tavern.py:298-321` (`test_gamble_loss`) has the same shape.

Replacing `Tavern.gamble` with a method that raises, then running the four gamble tests:

```
test_gamble_win_shows_correct_amount: PASSED with gamble() removed  <-- tests nothing
test_gamble_loss: PASSED with gamble() removed  <-- tests nothing
test_gamble_win_pays_multiple_of_bet: correctly failed -> gamble() was called
test_gamble_loss_via_real_loop: correctly failed -> gamble() was called
```

## Why it matters

Beyond testing nothing, the hand-written arithmetic is **wrong**. `test_gamble_win_shows_correct_amount` pays even money (`money += currentBet`, asserting `money == 150` on a $50 bet), while `src/location/tavern.py:238` pays five times the stake:

```python
winAmount = self.currentBet * DICE_WIN_MULTIPLIER
```

with `DICE_WIN_MULTIPLIER = 5` (`src/location/tavern.py:20`). The real payout on that setup is $350, not $150.

So the suite contains a test named "shows correct amount" that asserts the payout the game stopped using when #126 was fixed, and it cannot fail — a future change to `DICE_WIN_MULTIPLIER`, or to the win branch entirely, leaves it green. It reads as coverage of the gambling payout while providing none, which is worse than an acknowledged gap.

[...]

Issue #153, opened 2026-08-02
The server-backed web front-end shows "lost connection" instead of the ended screen when the game finishes


dmccoystephenson:
## Summary

During a triage pass, `cleanup()` was found to have exactly one production caller, on the Pyodide path only.

- `src/ui/webUserInterface.py:305-310` — the only thing that publishes the ended screen and stops the server:
  ```python
  def cleanup(self):
      self._present({"type": "ended"})
      if self._server is not None:
          self._server.shutdown()
          self._server.server_close()
  ```

- `web/pyodide_main.py:32-35` — the sole caller, whose own comment records the gap:
  ```python
  # Nothing else calls cleanup() when the game loop ends, and without it the
  # tab would sit on the last screen forever rather than saying it's over.
  if game is not None:
      game.userInterface.cleanup()
  ```

- `grep -rn "cleanup" src/ web/ examples/ --include='*.py'` returns only the three definitions, pygame's five internal QUIT handlers, and that one call. `FishE.play()` (`src/fishE.py:207-254`) returns without it, and `examples/web_app.py:40-41` is just `game = FishE(interfaceType=UIType.WEB); game.play()`.

## Why it matters

Under `UIType.WEB`, retiring or quitting ends `play()`, `main()` returns, the process exits and the daemon HTTP thread dies. The browser never receives an `ended` screen, so `poll()` accumulates failures and takes the disconnect branch (`src/ui/webUserInterface.py:94-111`), showing "Lost connection to the game — is it still running? Retrying…".

A player who has just reached the $10,000 goal and retired is told something broke.

The page even carries a guard for a state it can never reach on this front-end (`src/ui/webUserInterface.py:104-107`):
```js
// Don't clobber the intentional "game ended" screen with a scary banner.
const screen = FisheClient.getCurrentScreen();
if (failures === 5 && !(screen && screen.type === "ended")) {
```

The same gap applies to `src/fishE.py:351-353` under `UIType.PYGAME`, where `pygame.quit()` is never reached on a normal exit, and to `src/fishE.py:179` (`exit(0)` from the save-file menu), which bypasses `cleanup()` for every front-end.

This is the repo's recurring defect shape — a fix landing in one front-end only. It is worth fixing where all four benefit rather than adding a second copy of the Pyodide call.

## Suggested fix

Call it once, centrally, so no front-end can be forgotten. In `FishE.play()`, wrap the loop:

```python
def play(self):
    try:
        while self.running:
            ...
    finally:
[...]

Issue #152, opened 2026-08-02
Crew say the fleet is full and to buy a bigger boat while berths are still free on the other boats


dmccoystephenson:
## Summary

During a triage pass, the crew's "we're full" dialogue and its trigger condition were found to measure a single hull's berths, while hiring is capped fleet-wide.

- `src/npc/villagers.py:201-207` — the response, and `:247-248` — the condition that gates it, both use the best hull's `maxWorkers`:
  ```python
  def crowdedDialogue():
      info = business.tierInfo(business.currentTier(player))
      return (
          "All %d berths full on the %s. Elbow to elbow out there! If you "
          "want more hands you'll need more boat."
          % (info["maxWorkers"], info["name"])
      )
  ```
  ```python
  "condition": lambda: player.workers
  >= business.tierInfo(business.currentTier(player))["maxWorkers"],
  ```
  `business.currentTier(player)` is the best hull in the fleet (`src/business/business.py:57`), not a total.

- `src/business/boats.py:301-310` — the actual rule is the sum across every boat owned, and the docstring says so explicitly:
  ```python
  Hiring is now fleet-wide rather than per-boat: the cap is the total berths
  across every boat owned ...
  if player.workers >= totalCrewBerths(player):
      return False
  ```
  with `totalCrewBerths` = `sum(maxCrew(boat) for boat in player.boats)` (`src/business/boats.py:175-177`).

Run against the real modules with two Rowboats:

```
best hull: Rowboat -> maxWorkers = 5
fleet-wide totalCrewBerths = 10 (2 boats)

at workers=5:
  crowdedDialogue condition fires ('need more boat')? True
  yet hireWorker() still succeeds? True -> workers now 6
```

## Why it matters

The dialogue is not merely stale flavour text — it gives a specific, wrong purchasing instruction. A player with two Rowboats is told at 5 hands that every berth is full and that more crew requires a bigger boat, when 5 berths are free and hiring works normally. Acting on the advice means spending on an upgrade that buys nothing they did not already have.

`README.md:60` states the correct rule ("Crew come from one shared roster"), so the game documents fleet-wide hiring in the README and contradicts it in the mouth of the crew.

## Suggested fix

Point both the condition and the text at the same function hiring uses:

```python
def crowdedDialogue():
    return (
        "All %d berths full across the fleet. Elbow to elbow out there! If "
        "you want more hands you'll need another boat."
        % boats.totalCrewBerths(player)
    )
```
```python
[...]

Issue #151, opened 2026-08-02
The overnight fleet report is discarded at three of the six places a day rolls over


dmccoystephenson:
## Summary

During a triage pass, the `report` half of `increaseDay()`'s return value was found to be dropped at half its call sites, while the sibling `evicted` field is propagated.

`src/world/timeService.py:46-75` returns `{"evicted": bool, "report": [str]}`, and its own docstring states the contract:

> Returns `{"evicted": bool, "report": [str]}`. The report is what the fleet did overnight (see boats.describeDay) - now that every role earns, and a pirate crew can come home a man short, **the player has to be told rather than left to notice their roster changed.**

`README.md:65` makes the same promise to the player:

> You get an **overnight report** when a day passes telling you what the fleet did: what it landed, what it took, and — because a pirate crew doesn't always come home whole — who you lost.

**Sites that honour it (3):**
- `src/location/home.py:80-90` — sleeping
- `src/location/tavern.py:216-220` — drinking
- `src/location/docks.py:1031` — exporting, via the `_reportTheDay` helper

**Sites that drop it (3):**
- `src/location/docks.py:878-879` — every leg of a captained voyage. The summary is not even bound:
  ```python
  # A day at sea is a day at home too.
  self.timeService.increaseDay()
  ```
- `src/location/docks.py:1142` — a fishing trip long enough to cross 8am: `if self.timeService.increaseTime()["evicted"]:`
- `src/fishE.py:251` — the main loop's hourly tick, same shape.

`increaseTime()` returns `increaseDay()`'s dict verbatim when the clock reaches 8 (`src/world/timeService.py:42-43`), so the report is present at all three and simply not read.

## Why it matters

The voyage case also loses the eviction notice, not just the fleet lines. Each leg runs rent, wages, crew walkouts and possible eviction through `housing.runDailyRent`, and `docks.py:879` binds nothing — so a player can sail a five-day voyage and come home homeless without a single line of text saying so. `Docks` already owns the helper that would report it (`_reportTheDay`, `src/location/docks.py:928-935`), one method above the call.

For the other two, a day that rolls over while fishing or during the ordinary hourly tick tells the player nothing about what the fleet landed, what a raid cost, or which named crew member did not come back — precisely the outcome the docstring says must not happen.

## Suggested fix

[...]

PR #155, opened 2026-08-02
Fix README's no-dependencies claim and restore run.sh dependency install

### LINKED ISSUE #146: docs: README claims the web app needs no extra dependencies, but nothing outside the Dockerfile installs jsonschema
## Summary

`README.md:17` states, of running the game in the browser:

> The entire game — save-file manager, fishing, shop, bank, tavern, and NPC dialogue — plays in the browser, **with no extra dependencies (it uses only the Python standard library)**.

`jsonschema` is a hard, module-scope dependency of the game, not an optional one:

- `requirements.txt:8` — `jsonschema>=4.0,<5.0`
- `src/fishE.py:3` — `from jsonschema.exceptions import ValidationError`
- `src/validation/schemaValidator.py:4` — `from jsonschema import validate`

So `python3 examples/web_app.py` — the exact command the README gives two lines above that claim — raises `ModuleNotFoundError` on a clean interpreter. The Dockerfile is aware of this and says so in its own comment ("the game cannot start without it", `Dockerfile:5-8`); the claim is true only of `WebUserInterface`'s *implementation*, which uses `http.server`.

Nothing outside the Dockerfile installs it either:

- `install.sh` clones the repo and writes a launcher; it never runs `pip` at all.
- `run.sh` would have — but its `checkDependencies()`, the function containing `pip install -r requirements.txt`, is **commented out** at both its definition (`run.sh:23-39`) and its call site (`run.sh:58`).
- `README.md` has no installation section; `## Features` is the first heading and `## Contributing` the last.

## Why it matters

Every documented source-install path produces a game that cannot start, and the README actively asserts that no install step is needed. Only the container path works, and only by accident of it being the one place the dependency is handled.

## Suggested fix

1. Correct `README.md:17` — scope the standard-library claim to the web front-end's implementation, not to the game's dependencies.
2. Add a short install section to the README covering `pip install -r requirements.txt` before either entry point.
3. Either uncomment and repair `checkDependencies()` in `run.sh`, or have `install.sh` install `requirements.txt` after cloning. (While there: `runTests()` at `run.sh:41-46` discards pytest's exit status, so the game starts regardless of the result, and `pytest` is not a declared dependency either.)

dmccoystephenson:
## Summary
- README claimed both browser front-ends "use only the Python standard library," but `jsonschema` is a hard, module-scope dependency of the game itself (`src/fishE.py`, `src/validation/schemaValidator.py`) — every documented source-install path actually raises `ModuleNotFoundError` on a clean interpreter.
- Added a `## Setup` section to `README.md` with the `pip install -r requirements.txt` step, and corrected the "Play in your browser" claim to scope which front-end needs a local install (server-backed web) vs. an automatic in-browser one (Pyodide).
- Uncommented and repaired `checkDependencies()` in `run.sh` (checks `python3` instead of `python`, exits nonzero on failure, installs `pytest` alongside `requirements.txt`) and re-enabled its call in `main`.
- Made `runTests()` abort with the test's exit code instead of always falling through to `startProgram()` regardless of test results.

## Deferred backlog (recorded per Phase 1)
This cycle picked #146 as the sole issue (doc/build-fix tiebreaker). Other open issues were left for later cycles:
- Save-file-safety cluster (#149, #150, #143, #142) — touches protected paths (`src/saveFileManager.py`, `*JsonReaderWriter.py`) that need an explicit human merge decision per the do-not-auto-merge gate, so sequencing them after clean work avoids stranding the loop.
- Dialogue-accuracy bugs (#144, #145, #152) and the test/config cleanups (#154, #147, #151, #153) are all clean, non-protected candidates for upcoming cycles.

## Test plan
- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 720 passed
- [x] `bash -n run.sh` — syntax OK

Closes #146

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,14 @@
 
 This game allows you to explore a fishing village and perform actions in it.
 
+## Setup
+
+```bash
+pip install -r requirements.txt
+```
+
+The game itself depends on `jsonschema` (it validates save files on every load and save), so this step is required before running any front-end from a local Python interpreter — console or the server-backed web front-end (`python3 examples/web_app.py`). `run.sh` runs this for you. The pygame front-end (`UIType.PYGAME`) additionally needs `pip install pygame`, kept out of `requirements.txt` since it's only imported lazily when that front-end is selected.
+
 ## Features
 
 ### Play in your browser
@@ -26,7 +34,7 @@ python3 examples/web_app.py
 # then open http://127.0.0.1:8000
 ```
 
-Both use only the Python standard library; the Pyodide front-end loads its one runtime dependency (`jsonschema`) in the browser.
+Both need `jsonschema` (see Setup above) — the Pyodide front-end loads it into the browser sandbox automatically when the page opens, while the server-backed front-end runs on your local Python and needs it installed first.
 
 ### One Thing at a Time
 A new game opens on the docks with a rod, a bucket, and exactly one thing to do: **fish**. Everything else in the village arrives later, one option at a time, as you earn it — and each arrival tells you why it's there:
--- a/run.sh
+++ b/run.sh
@@ -20,29 +20,35 @@ printVersion() {
     echo ""
 }
 
-# checkDependencies() {
-#     # check that dependencies are installed
-#     echo "Checking dependencies"
-#     if ! command -v python &> /dev/null
-#     then
-#         echo "Python could not be found. Download it from https://www.python.org/downloads/"
-#         exit
-#     fi
-#     if ! command -v pip &> /dev/null
-#     then
-#         echo "Pip could not be found. Download it from https://pip.pypa.io/en/stable/installation/ or run 'python -m ensurepip' in a terminal"
-#         exit
-#     fi
-#     pip install pygame --pre --quiet
-#     pip install -r requirements.txt --quiet
-#     echo ""
-# }
+checkDependencies() {
+    # check that dependencies are installed
+    echo "Checking dependencies"
+    if ! command -v python3 &> /dev/null
+    then
+        echo "Python could not be found. Download it from https://www.python.org/downloads/"
+        exit 1
+    fi
+    if ! command -v pip &> /dev/null
+    then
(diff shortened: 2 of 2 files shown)


dmccoystephenson:
Self-review rubric:
- Scope: PASS — every changed line (README.md Setup section + Play-in-your-browser paragraph; run.sh checkDependencies/runTests/main) is necessary for #146; no unrelated files touched.
- Tests-new: N/A — this PR changes documentation and a shell launcher script, not a public Python method/function; there is no pytest surface to add coverage to. Verified `bash -n run.sh` for syntax instead.
- Tests-fix: N/A — not a src/ bug fix with a regression-test contract; verified behavior by re-reading the diff (checkDependencies now runs unconditionally in `main`, runTests now `exit`s nonzero on pytest failure before reaching startProgram).
- Sibling structure: N/A — no new files created.
- Sibling renames: N/A — no renamed identifiers.
- Docs: PASS — README.md's Setup section and the "Play in your browser" paragraph now match actual behavior (jsonschema is a hard runtime dependency for both Python-hosted front-ends; only Pyodide loads it automatically in-browser).
- Issue resolution: PASS — all three of #146's suggested-fix items are done: (1) README's standard-library claim corrected, (2) a Setup section with `pip install -r requirements.txt` added, (3) run.sh's checkDependencies() uncommented/repaired and re-enabled, plus the noted pytest-exit-status gap fixed and pytest added to the install step.
- CI: PASS — `test` check green on the PR head SHA (720 tests passed).

Repo-specific items (Schema-sync, Money-format, Deterministic-tests, Headless-pygame, camelCase): N/A — no persisted fields, money display, tests, or Python method names touched by this PR.

Out-of-diff observation, not introduced by this PR: `checkDependencies()` still checks `command -v pip` rather than `pip3`; on a host where only `pip3` exists this would report "Pip could not be found" even though a working pip is present. That predates this change (present in the original commented-out block) and is outside #146's scope, so left as-is rather than folded in here.

---

Issue #51, opened 2026-02-05
Multiple save files


dmccoystephenson:
As a user, I would appreciate the ability to have multiple save files and be able to detect/manage these saves when the application starts.

PR #165, opened 2026-08-02
Remove four Config settings nothing reads

### LINKED ISSUE #147: Config declares four settings nothing reads, and the tests assert they keep existing
## Summary

Four of `Config`'s ten attributes are never read anywhere in `src/`:

- `src/config/config.py:6-8` — `playerSaveFile`, `statsSaveFile`, `timeServiceSaveFile`. Save paths are resolved through `SaveFileManager.get_save_path()` instead (`src/fishE.py:283`, `:290`, `:297`, `:307`, `:320`, `:331`), which composes them from the selected slot. A grep for these names outside `config.py` matches only *local variables* in `fishE.py` (`as playerSaveFile`), never `self.config.playerSaveFile`.
- `src/config/config.py:12` — `initialEnergy`. `Player` deliberately does not use it, and says why: starting energy comes from the housing ladder instead, since a flat configured value could exceed the Homeless cap (`src/player/player.py:14-19`). The only mention of the name outside `config.py` is that explanatory comment.

The remaining six (`dataDirectory`, `initialMoney`, `initialFishCount`, `initialMoneyInBank`, `initialFishMultiplier`, `initialPriceForBait`) are genuinely wired in, so this is dead surface on an otherwise live class.

`tests/config/test_config.py:14-16` and `:20` assert the dead values, which means the test suite currently pins configuration that changing would break nothing but the tests.

## Why it matters

Minor, but actively misleading: the three save-path settings look like the way to relocate save files, and editing them silently does nothing — the multi-slot layout introduced in #51 made them obsolete. `initialEnergy` is a trap in the opposite direction, since honouring it would reintroduce the bug `player.py` documents.

## Suggested fix

Delete the three save-path attributes and their assertions; the slot layout is `SaveFileManager`'s responsibility and `dataDirectory` already covers the configurable part. For `initialEnergy`, either delete it too or add a comment at the declaration pointing at `player.py`'s reasoning, so the next reader does not "fix" the fact that it is unused.

dmccoystephenson:
## Summary
- Removed `Config.playerSaveFile`, `statsSaveFile`, and `timeServiceSaveFile` — the multi-slot save layout resolves paths through `SaveFileManager.get_save_path()` instead, so these three have been dead since #51; a grep confirms every remaining match outside `config.py`/the test was a local `as` variable in `fishE.py`, never `self.config.*`.
- Removed `Config.initialEnergy` — `player.py` already documents that starting energy deliberately comes from the housing ladder, not a configured value, so the attribute had no reader. Updated that comment since it referenced the now-removed attribute by name.
- Trimmed the matching assertions out of `tests/config/test_config.py` (they were pinning dead values) and dropped the now-unused `os` import.

## Test plan
- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest` — 729 passed
- [x] `grep -rn '\.playerSaveFile\|\.statsSaveFile\|\.timeServiceSaveFile\|\.initialEnergy' --include='*.py' .` — no remaining references anywhere
- [x] README/PLANNING.md grepped for the removed names — no doc drift

Closes #147

---

### DIFF
--- a/src/config/config.py
+++ b/src/config/config.py
@@ -10,13 +10,10 @@ class Config:
         # Worker-side directory that the Pyodide front-end mirrors to the
         # browser's IndexedDB (see browserSaveSync and web/game-worker.js).
         self.dataDirectory = os.environ.get("FISHE_SAVE_DIR") or "data"
-        self.playerSaveFile = os.path.join(self.dataDirectory, "player.json")
-        self.statsSaveFile = os.path.join(self.dataDirectory, "stats.json")
-        self.timeServiceSaveFile = os.path.join(self.dataDirectory, "timeService.json")
 
-        # Initial player values
+        # Initial player values. Starting energy isn't configured here - see
+        # player.py, which sources it from the housing ladder instead.
         self.initialMoney = 20
-        self.initialEnergy = 100
         self.initialFishCount = 0
         self.initialMoneyInBank = 0.01
         self.initialFishMultiplier = 1
--- a/src/player/player.py
+++ b/src/player/player.py
@@ -12,10 +12,10 @@ class Player:
         self.fishMultiplier = 1 if config is None else config.initialFishMultiplier
         self.priceForBait = 50 if config is None else config.initialPriceForBait
         # Starts at the Homeless tier's energy cap (see src/housing) - a
-        # fresh player hasn't found anywhere to stay yet. Deliberately not
-        # sourced from Config.initialEnergy: the housing ladder is the
-        # source of truth for energy caps (see src/housing/housing.py), and
-        # a flat configured starting energy could exceed the Homeless cap.
+        # fresh player hasn't found anywhere to stay yet. Deliberately not a
+        # configured value: the housing ladder is the source of truth for
+        # energy caps (see src/housing/housing.py), and a flat configured
+        # starting energy could exceed the Homeless cap.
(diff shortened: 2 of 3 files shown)


dmccoystephenson:
Self-review rubric:
- Scope: PASS — only the three files needed to remove the four dead settings and update the one comment that named them.
- Tests-new: N/A — this removes dead surface rather than adding behavior; nothing new needs its own test.
- Tests-fix: N/A — not a bug fix.
- Sibling structure: N/A — no new files.
- Sibling renames: N/A — no renames.
- Docs: PASS — README.md/PLANNING.md grepped for the four removed names, no hits.
- Issue resolution: PASS — all four settings named in #147 (playerSaveFile, statsSaveFile, timeServiceSaveFile, initialEnergy) are removed, along with their test assertions, per the issue's suggested fix.
- CI: PASS — test check green on f6b7ae0 (https://github.com/Stephenson-Software/FishE/actions/runs/30771347132).
- Schema-sync: N/A — Config isn't a persisted model.
- Money-format: N/A.
- Deterministic-tests: N/A — no test touches random.
- Headless-pygame: N/A.
- camelCase: N/A — no new identifiers.

Summary: dead-code removal, all applicable items pass, CI green. Ready to merge.

---

Issue #156, opened 2026-08-02
test: 30 voyage-event handlers are asserted only to return a non-empty string, so their payouts are unverified


dmccoystephenson:
## Summary

During a triage pass, the voyage-event outcome handlers in `src/business/adventures.py` were found to have narration-only coverage. Every handler mutates voyage state — `gain(voyage, money=…, fishCount=…)`, `damage()`, `repair()`, `useSupplies()`, `loseCrew()` — and no test asserts any of those effects.

The only two tests that reach them check the return value is a truthy string:

- `tests/business/test_adventures.py:104-112` (`test_every_choice_returns_narration`):
  ```python
  text = adventures.resolveChoice(voyage, choice)
  assert isinstance(text, str) and text
  ```
- `tests/business/test_adventures.py:313-328` (`test_every_outcome_branch_is_reachable`):
  ```python
  assert adventures.resolveChoice(voyage, choice)
  ```
  which also resets `hull`, `status` and `crew` between choices, so even incidental state is discarded.

No test references any handler by name:
```
$ grep -rn "merchantBoard\|patrolFight\|goodGroundsNets\|leakShipwright\|squallRunBefore\|merchantShadow" tests/
(zero matches)
```

Demonstrated by mutation. `_merchantShadow` (`src/business/adventures.py:358-364`) pays $400-900, the largest single payout in the game. Changing line 360 from `gain(voyage, money=money)` to `gain(voyage, money=0)`:

```
720 passed in 45.28s
```

The whole suite is green with that payout deleted.

## Why it matters

These handlers carry the biggest numbers in the economy — `_merchantShadow` $400-900 and `_patrolFight` $300-800 against a passive fleet rate of $14-26 per crew per day. A sign flip, a wrong keyword in `gain()`, a `repair` where `damage` was meant, or a `loseCrew` that removes the wrong hand would all ship green.

The existing tests read as thorough — one iterates every choice of every event for every role, the other forces the dice to both extremes — which makes the gap easy to miss: the suite proves every branch is *reachable* and proves nothing about what any of them *do*.

## Suggested fix

Add state assertions rather than replacing the sweeps, which are valuable as reachability checks. Highest value first, since these move money or crew:

- `_merchantBoard` (`:337`), `_merchantShadow` (`:358`), `_patrolFight` (`:383`) — assert `voyage["money"]` lands in the documented range
- `_goodGroundsFish` (`:315`), `_goodGroundsNets` (`:321`) — assert `fishCount`
[...]

Issue #150, opened 2026-08-02
A save whose player.json is corrupt disappears from the menu and its slot is offered as "Create New Save"


dmccoystephenson:
## Summary

During a triage pass, `_read_save_metadata` was found to return `None` for an unreadable `player.json`, which removes the whole slot from the save list rather than reporting it.

- `src/saveFileManager.py:92-94` — the exception is bound and never used; no log, no re-raise, no signal to any front-end:
  ```python
  except (json.JSONDecodeError, IOError, OSError) as e:
      # Return None for corrupted or inaccessible save files
      return None
  ```

- `src/saveFileManager.py:44-45` — a slot is only listed `if metadata:`, so a `None` drops it silently.

- `src/saveFileManager.py:96-108` — `get_next_available_slot()` derives `existing_slots` from that same filtered list, so it hands back the number of the slot that is still very much occupied.

The result, run against the real `SaveFileManager` with a slot holding a truncated `player.json` next to an intact `stats.json` and `timeService.json`:

```
list_save_files() -> []
get_next_available_slot() -> 1
slot_1 still on disk -> ['player.json', 'stats.json', 'timeService.json']
```

`_selectSaveFile` (`src/fishE.py:157-160`) therefore renders `Create New Save (Slot 1)` and never renders a `Load Slot 1` entry.

## Why it matters

A single corrupt byte in `player.json` — from a crash or a kill during the truncate-then-write in `save()` (`src/fishE.py:282-300`), which runs after **every** player action — makes a long-running save vanish from the menu with no message.

The player is not told a save existed, so they have no reason to copy the directory off before continuing. Picking the only offered option points `select_save_slot(1)` at the existing directory, and the next `save()` overwrites the intact `stats.json` and `timeService.json` sitting beside the damaged file. Data that was recoverable by hand is destroyed by the act of starting a new game.

This is distinct from #143: there, a load is attempted and fails. Here no load is attempted at all — the slot is *invisible*, and the reported symptom is "my save is gone" rather than "my save reset".

## Suggested fix

Do not let an unreadable slot vanish. Return a marker instead of `None` so the slot stays in the list and stays claimed, e.g. `{"corrupt": True, "last_modified": ...}`, then:

- render it in the menu as something like `Slot 1 (unreadable - will not be overwritten)`, non-selectable for loading;
- have `get_next_available_slot()` count it as occupied so a new game lands in slot 2;
[...]

Issue #143, opened 2026-08-01
A save file that fails to load is silently replaced with a new game and overwritten on the next action


dmccoystephenson:
## Summary

When a save file cannot be read, the load handlers fall back to fresh objects and the game then proceeds to write those fresh objects straight over the file that failed:

- `src/fishE.py:313-315` — a corrupt `player.json` is caught and replaced with `self.player = Player(self.config)`
- `src/fishE.py:324-326`, `src/fishE.py:339-341` — same for `stats.json` and `timeService.json`
- `src/fishE.py:253` — `self.save()` runs at the end of **every** iteration of the game loop
- `src/fishE.py:283`, `:290`, `:297` — `save()` opens each path with mode `"w"`, which truncates the existing file before anything is written

So one unreadable file costs the player their entire slot: it loads as a brand-new character, and the first action taken overwrites the original bytes. Because the only notice is a `print()` (see #142), nothing on screen distinguishes this from a normal start.

The truncating write is also the likely way a slot gets into that state in the first place — `"w"` empties the file immediately, so an exception or a kill between the truncate and the completed `json.dump` leaves a partial file on disk with no intact copy anywhere.

This is adjacent to but distinct from #57 (KeyError escaping the handlers) and #111 (schema validation of loaded data): both made the *detection* of a bad save robust; neither changed what happens to the file afterwards.

## Why it matters

This is unrecoverable loss of the player's whole run, from a failure mode the player is never shown and did not cause. It is worse for the Docker/web deployment, where saves live on a named volume (`fishe-data`) and there is no user-accessible copy to restore from.

## Suggested fix

Two independent changes, either of which alone reduces the blast radius:

1. **Write atomically.** Serialize to a temporary file in the same directory and `os.replace()` it over the target. `os.replace` is atomic on POSIX and Windows, so a crash mid-write leaves the previous good save intact instead of a truncated one.
2. **Do not overwrite a slot whose load failed.** Track that a load fell back, and either refuse to save over that slot or move the unreadable file aside (e.g. `player.json.corrupt`) before writing, so the original bytes survive for manual recovery.

_Filed by Claude during an automated triage pass; claims above were verified against source._

Issue #142, opened 2026-08-01
Save and load failures are reported with print(), so no front-end ever shows them


dmccoystephenson:
## Summary

Every save/load error path in `fishE.py` reports itself with `print()` rather than through the active front-end, so the warning reaches no player on any of the three interfaces.

- `src/fishE.py:301` — `print(f"\n Warning: Failed to save game: {e}")`
- `src/fishE.py:313-314` — `print(f"\n Warning: Failed to load player data: {e}")` / `print(" Creating new player...")`
- `src/fishE.py:324-325`, `src/fishE.py:339-340` — the same pattern for stats and the time service

On the pygame and web front-ends nothing written to stdout is rendered at all — it goes to the terminal (or the container log) that the player is not looking at. On the console front-end it is scrolled away immediately: the very next screen calls `lotsOfSpace()`, which prints 20 blank lines (`src/ui/userInterface.py:22`, called first thing in `showOptions` at `src/ui/userInterface.py:35`).

This is the same class of defect as #125, which was fixed for the fishing/drinking flavour text; these paths were not converted.

## Why it matters

The failure being hidden is precisely the failure a player most needs to know about. A save that did not write, or a save that did not load, is silently indistinguishable from a normal turn — see #143 for what happens next after a failed load.

## Suggested fix

Route these through the front-end contract rather than stdout. `BaseUserInterface.showDialogue` is implemented by all three front-ends and waits for acknowledgement, which is the right weight for "your game did not save". `self.userInterface` is already available on `FishE` at every one of these call sites; note that `loadPlayer`/`loadStats`/`loadTimeService` run during `__init__` after the UI is built (`src/fishE.py:45`), so it is in place by then.

_Filed by Claude during an automated triage pass; claims above were verified against source._

PR #167, opened 2026-08-03
Reject non-positive bank amounts and fix stale crew/interest dialogue

### LINKED ISSUE #149: A negative bank deposit prints money and writes a save the game will refuse to load
## Summary

During a triage pass, the deposit and withdraw handlers were found to accept negative amounts. Both the sign check and the schema check are missing on the write side, so the resulting save is rejected on the next launch.

- `src/location/bank.py:226-228` — the guard is an affordability test, which any negative amount passes:
  ```python
  if self.player.canAfford(amount):
      self.player.moneyInBank += amount
      self.player.spendMoney(amount)
  ```
  `canAfford` is `self.money >= cost` (`src/player/player.py:78-79`), and `-100 <= money` is always true. `spendMoney` then does `self.money -= cost` (`src/player/player.py:81-83`), so a negative cost *adds* cash.

- `src/location/bank.py:244-246` — `withdraw` has the mirror-image hole: `if amount <= self.player.moneyInBank:` passes for any negative, driving `player.money` negative.

- Nothing filters the input. `promptForNumber` is `float(self.promptForText(...))` (`src/ui/baseUserInterface.py:103-109`), and the browser client only checks `!isNaN(Number(inp.value))` (`web/client.js:99-100`). `-100` is a valid float in all four front-ends.

Run against the real classes:

```
start: money=20.00 bank=0.01
after deposit(-100): money=120.00 bank=-99.99  -> $-100.00 deposited successfully.
after withdraw(-500): money=-380.00 bank=400.01 -> $-500.00 withdrawn successfully.
```

## Why it matters

Two separate harms.

**The economy is trivially bypassed.** Repeating `deposit -100` mints unlimited cash, which makes the $10,000 goal, the housing ladder and the whole fleet progression skippable from the first visit to the bank.

**It destroys the save on the next launch.** `schemas/player.json` declares `"minimum": 0` for both `money` and `moneyInBank`, but validation runs only on the read side (`src/player/playerJsonReaderWriter.py:64`), never on write (`:100-102`). So `save()` writes the file happily, and the next load raises:

```
WRITE: succeeded, bytes=283
READ: ValidationError: -99.99 is less than the minimum of 0
```

That `ValidationError` is caught at `src/fishE.py:319` and turned into a brand-new player — so one mistyped `-` at the deposit prompt silently costs the player their entire run, with the loss surfacing a session later than the typo.

Secondary effect: negative `player.money` makes `boats.runDailyProduction` compute `affordable = min(player.workers, int(player.money // 10))` as negative (`src/business/boats.py:442-446`), shedding the whole crew.

## Suggested fix

Reject non-positive amounts in both handlers, next to the existing `is None` retry branch:

```python
if amount is None or amount <= 0:
    self.currentPrompt.text = "Enter an amount greater than zero. Money: $%.2f" % self.player.money
    continue
```

Separately, and more durably: call `validate_against_schema` on the write side of the three `*JsonReaderWriter` classes as well as the read side. The schema already encodes the invariant; validating only on read means the game can produce files it will not accept, and the report arrives one launch too late to save the data.

### LINKED ISSUE #152: Crew say the fleet is full and to buy a bigger boat while berths are still free on the other boats

### LINKED ISSUE #158: The bank teller says the more you save the more you earn, but interest is capped at $50/day and the cap is never disclosed
## Summary

During a triage pass, the bank's dialogue was found to state a rule the interest calculation stopped following when the #66 exploit was capped.

- `src/location/bank.py:54` — Margaret's answer:
  > "The more you save, the more interest you earn!"

- `src/location/bank.py:45` — the same claim in her opening pitch:
  > "Plus, your savings earn interest over time. The more you save, the more you earn. "

- `src/world/timeService.py:13-14` and `:57-58` — the actual rule:
  ```python
  INTEREST_RATE = 0.02
  MAX_INTEREST_PER_DAY = 50
  ...
  moneyToAdd = int(math.ceil(self.player.moneyInBank * INTEREST_RATE))
  moneyToAdd = min(moneyToAdd, MAX_INTEREST_PER_DAY)
  ```

At 2%, the cap binds from $2,500 banked. Past that, every additional dollar deposited earns exactly nothing.

## Why it matters

The wealth goal is $10,000 (`GOAL_AMOUNT`, `src/achievements/achievements.py`), so the cap binds across most of a run — the stated rule is false for the majority of the time a player is likely to be using the bank.

It also misinforms a real decision. A player told "the more you save, the more you earn" has a reason to bank a windfall rather than spend it on a boat or a house. Past $2,500 that is simply wrong: the marginal return on banking is zero, and the money is strictly better deployed in the fleet or the housing ladder. Nothing in-game or in the README mentions a cap, so there is no way for the player to discover this except by noticing the number stops moving.

This is not a re-run of #66, which was about the uncapped 10% compounding exploit and is fixed. The cap works; the dialogue was never updated to match it.

## Suggested fix

Make the teller state the actual rule, including the ceiling, and derive the numbers from the constants so the line cannot drift again:

```python
"Your savings earn %d%% a day, up to $%d — past about $%d banked you've "
"maxed out what I can pay you." % (
    INTEREST_RATE * 100, MAX_INTEREST_PER_DAY,
    int(MAX_INTEREST_PER_DAY / INTEREST_RATE),
)
```

Both lines (`:45` and `:54`) need it. A player who understands the ceiling can plan around it, which makes the cap a design feature rather than something that quietly stops working.

### LINKED ISSUE #160: Write saves atomically and validate on write, so a crash or a bad value can't destroy a run
## Summary

During a triage pass, two structural weaknesses in the save path were found underneath the specific bugs filed as #149 and #150. Both are the same shape: the game can produce a save file it will later refuse to read, and it destroys the previous good copy in the act of producing it.

**Writes are not atomic.** `src/fishE.py:282-300` opens each of the three files with mode `"w"`, which truncates the existing file to zero bytes before anything is written:

```python
try:
    with open(
        self.saveFileManager.get_save_path("player.json"), "w"
    ) as playerSaveFile:
        self.playerJsonReaderWriter.writePlayerToFile(
            self.player, playerSaveFile
        )
```

There is no temp-file-and-rename anywhere in the repo. `save()` runs at the end of every loop iteration (`src/fishE.py:254`) — after every single player action — so this window is entered dozens of times a session. A Ctrl-C (not in the `except (IOError, OSError)` tuple), a closed terminal, a stopped container or a full disk during that window leaves a truncated `player.json`.

The three writes are also not a unit: if `player.json` succeeds and `stats.json` then fails, the `return` at `src/fishE.py:304` leaves the slot internally inconsistent — and, on the Pyodide front-end, skips `syncBrowserSaves()` at `:309` entirely, so even the successful write never reaches IndexedDB.

**Validation runs on read only.** `src/player/playerJsonReaderWriter.py:64` validates against the schema when loading, but `writePlayerToFile` (`:100-102`) does not. Same asymmetry in `src/stats/statsJsonReaderWriter.py:126-128` and `src/world/timeServiceJsonReaderWriter.py:32-34`. So an invariant the schema already declares — `schemas/player.json` requires `"minimum": 0` on `money` and `moneyInBank` — is enforced one launch too late:

```
WRITE: succeeded, bytes=283
READ: ValidationError: -99.99 is less than the minimum of 0
```

That `ValidationError` is caught at `src/fishE.py:319` and turned into a fresh player.

## Why it matters

These two gaps are what turn ordinary mistakes into total losses. #149 is a missing sign check, but the reason it costs a player their run rather than producing a silly number is that the bad value is written unvalidated and only rejected at next launch. #150 is a missing error path, but the reason there is a corrupt file to trip over is that the write is non-atomic.

Fixing the two specific bugs leaves the mechanism intact for the next one. Any future field with a schema constraint has the same trapdoor.

## Suggested fix

**Atomic writes.** Write each file to a sibling temp and rename, which is atomic on POSIX and on Emscripten's filesystem:

```python
tmp = path + ".tmp"
with open(tmp, "w") as f:
    writer(obj, f)
os.replace(tmp, path)
```

Rename all three only after all three temps are written, so a slot never contains a mix of old and new. Move `syncBrowserSaves()` so it runs whenever any write succeeded, rather than being skipped by the early `return`.

**Validate on write.** Call `validate_against_schema` in each `write*ToFile`, as the read side already does. Failing there surfaces the bad value while the good state is still in memory and the old file is still intact — the difference between "that deposit was rejected" and "your save is gone".

Related: #149 and #150 are the two concrete failures this would have prevented; #111 (closed) added the read-side validation this proposes to mirror.

dmccoystephenson:
## Summary

Three small, independent bugs found during triage are addressed here.

- **#149** — Negative amounts were accepted by `Bank.deposit()`/`Bank.withdraw()`. A negative deposit passed `canAfford()` (money >= a negative number is always true) and minted cash, while a negative withdrawal drove `player.money` negative — both wrote a save that `schemas/player.json`'s `"minimum": 0` constraint would reject on the next load. Non-positive amounts are now rejected with a re-prompt, mirroring the existing `is None` retry branch.
- **#152** — The crew NPC's "berths full" dialogue and its unlock condition were keyed off the best hull's `maxWorkers`, even though hiring has been fleet-wide (`boats.totalCrewBerths`) since a prior change. A player with a second boat and open berths was told to buy a bigger boat while hiring still worked normally. Both the condition and the response text are now derived from `totalCrewBerths`, matching the README's documented "one shared roster" rule.
- **#158** — The bank teller's "interest rates" dialogue claimed "the more you save, the more you earn," which stopped being true once `MAX_INTEREST_PER_DAY` was introduced to cap the daily payout. The line is now derived from `INTEREST_RATE`/`MAX_INTEREST_PER_DAY` so it states the real rule and the point at which it caps, and can't drift from the constants again.

The broader, structural write-side validation this triage also flagged (unvalidated writes letting an invalid save land on disk at all) is left to #160, which is a materially larger change touching `saveFileManager.py`/`*JsonReaderWriter.py`.

## Issues deferred this cycle

The rest of the open backlog is deferred without further comment, since this cycle's scope was kept to three small, low-risk fixes: #160 (save-write atomicity/validation — large, touches protected save-path files), #156 (voyage-event test coverage — a sizeable standalone test-expansion effort), #153 (web front-end "ended" screen), #151 (overnight report dropped at three call sites), #150 (corrupt save vanishing from the menu), #143 and #142 (save/load error reporting).

## Test plan

- [x] Full suite: `731 passed` under headless SDL drivers (`SDL_VIDEODRIVER=dummy`, `SDL_AUDIODRIVER=dummy`)
- [x] New regression tests added for negative-amount rejection (deposit/withdraw), the interest-rate dialogue's disclosed cap, and the fleet-wide berths dialogue/condition

Closes #149
Closes #152
Closes #158

### DIFF
--- a/src/location/bank.py
+++ b/src/location/bank.py
@@ -1,13 +1,18 @@
 from location.enum.locationType import LocationType
 from player.player import Player
 from prompt.prompt import Prompt
-from world.timeService import TimeService
+from world.timeService import TimeService, INTEREST_RATE, MAX_INTEREST_PER_DAY
 from stats.stats import Stats
 from ui.userInterface import UserInterface
 from npc.npc import NPC
 from business import business
 from investments import investments
 from progression import progression
+
+# The break-even point past which the daily interest cap binds - used so the
+# teller's dialogue can state the real rule instead of drifting from it (see
+# src/world/timeService.py for the underlying constants).
+_INTEREST_CAP_BALANCE = int(MAX_INTEREST_PER_DAY / INTEREST_RATE)
 
 
 # @author Daniel McCoy Stephenson
@@ -42,16 +47,20 @@ class Bank:
                     "response": "The bank is simple and safe! You can deposit money when you have some on hand, "
                     "and withdraw it whenever you need. We keep your money secure - "
                     "no risk of losing it to gambling or spending it accidentally! "
-                    "Plus, your savings earn interest over time. The more you save, the more you earn. "
-                    "It's the smart way to grow your wealth!",
+                    "Plus, your savings earn interest over time - up to $%d a day. "
+                    "It's the smart way to grow your wealth!" % MAX_INTEREST_PER_DAY,
                 },
                 {
                     "question": "Tell me about interest rates.",
-                    "response": "Ah yes, interest! Every day that passes, your savings grow by a small percentage. "
-                    "It might not seem like much at first, but over time it really adds up! "
+                    "response": "Ah yes, interest! Every day that passes, your savings grow by %d%%, "
+                    "up to $%d a day - past about $%d banked you've maxed out what I can pay you. "
                     "The interest is automatically added to your bank account. "
-                    "Think of it as the bank paying you for keeping your money with us. "
-                    "The more you save, the more interest you earn!",
+                    "Think of it as the bank paying you for keeping your money with us."
+                    % (
+                        int(INTEREST_RATE * 100),
+                        MAX_INTEREST_PER_DAY,
+                        _INTEREST_CAP_BALANCE,
+                    ),
                 },
                 {
                     "question": "Should I save or spend my money?",
@@ -222,6 +231,12 @@ class Bank:
             if amount is None:
                 self.currentPrompt.text = "Try again. Money: $%.2f" % self.player.money
                 continue
+            if amount <= 0:
+                self.currentPrompt.text = (
+                    "Enter an amount greater than zero. Money: $%.2f"
+                    % self.player.money
+                )
+                continue
 
             if self.player.canAfford(amount):
                 self.player.moneyInBank += amount
@@ -238,6 +253,12 @@ class Bank:
             if amount is None:
                 self.currentPrompt.text = (
                     "Try again. Money In Bank: $%.2f" % self.player.moneyInBank
+                )
+                continue
+            if amount <= 0:
+                self.currentPrompt.text = (
+                    "Enter an amount greater than zero. Money In Bank: $%.2f"
+                    % self.player.moneyInBank
                 )
                 continue
 
--- a/src/npc/villagers.py
+++ b/src/npc/villagers.py
@@ -212,11 +212,10 @@ def createCrewNPC(player, name):
         )
 
     def crowdedDialogue():
-        info = business.tierInfo(business.currentTier(player))
         return (
-            "All %d berths full on the %s. Elbow to elbow out there! If you "
-            "want more hands you'll need more boat."
-            % (info["maxWorkers"], info["name"])
+            "All %d berths full across the fleet. Elbow to elbow out there! "
+            "If you want more hands you'll need another boat."
+            % boats.totalCrewBerths(player)
         )
 
     def businessNameDialogue():
@@ -257,8 +256,7 @@ def createCrewNPC(player, name):
                 # Only worth asking once there's no room left to hire.
                 "question": "Getting crowded out there, isn't it?",
                 "response": crowdedDialogue,
-                "condition": lambda: player.workers
-                >= business.tierInfo(business.currentTier(player))["maxWorkers"],
+                "condition": lambda: player.workers >= boats.totalCrewBerths(player),
             },
             {
                 # The outfit has to have a name before anyone can have an
--- a/tests/location/test_bank.py
+++ b/tests/location/test_bank.py
@@ -194,6 +194,26 @@ def test_npc_business_dialogue_fleet_tier():
     assert "retire a wealthy soul" in bankInstance._businessDialogue()
 
 
+def test_npc_interest_dialogue_discloses_the_daily_cap():
+    # prepare - the teller's "interest rates" answer must state the actual
+    # rule (the daily cap timeService.py enforces), not just "the more you
+    # save, the more you earn"
+    from src.world.timeService import INTEREST_RATE, MAX_INTEREST_PER_DAY
+
+    bankInstance = createBank()
+    options = bankInstance.npc.get_dialogue_options()
+    questions = [option["question"] for option in options]
+    index = questions.index("Tell me about interest rates.")
+
+    # call
+    response = bankInstance.npc.get_dialogue_response(index)
+
+    # check
+    assert "%d%%" % int(INTEREST_RATE * 100) in response
+    assert "$%d" % MAX_INTEREST_PER_DAY in response
+    assert "maxed out" in response
+
+
 def test_deposit_success():
     # prepare
     bankInstance = createBank()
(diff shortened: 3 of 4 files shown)


dmccoystephenson:
Self-review rubric:

- **Scope:** PASS — the diff touches only `src/location/bank.py`, `src/npc/villagers.py`, and their matching test files; every changed line belongs to one of #149/#152/#158.
- **Tests-new:** PASS — `test_deposit_negative_amount_is_rejected`, `test_withdraw_negative_amount_is_rejected`, `test_npc_interest_dialogue_discloses_the_daily_cap`, and `test_createCrewNPC_crowded_question_stays_hidden_with_room_on_another_boat` each exercise the new behavior.
- **Tests-fix (empirical):** PASS — source fixes were reverted to the `origin/main` state (tests left in place) and all four new tests were confirmed to FAIL (`4 failed`); the fix was then restored and the full suite passed (`731 passed`).
- **Sibling structure:** PASS — new tests follow the existing `# prepare`/`# call`/`# check` pattern used throughout `test_bank.py`/`test_villagers.py`.
- **Sibling renames:** N/A — no parallel identifier pair was renamed this PR.
- **Docs:** PASS — `README.md` already documents fleet-wide hiring ("Crew come from one shared roster") and says nothing about a specific interest rate/cap, so no doc update was needed; the code/dialogue now matches what was already documented.
- **Issue resolution:** PASS — #149 (deposit/withdraw now reject amounts <= 0), #152 (crowded dialogue + condition now use `totalCrewBerths`), #158 (interest dialogue now states the actual rate and cap) are each fully addressed by the diff, not partially.
- **CI:** PASS — `gh pr checks 167` is green (`test pass, 32s`).
- **Schema-sync:** N/A — no persisted field was added or changed.
- **Money-format:** PASS — the new prompts use the existing `%.2f` formatting already used elsewhere in `bank.py`.
- **Deterministic-tests:** PASS — no new test depends on unseeded `random`.
- **Headless-pygame:** N/A — no pygame-touching code was changed.
- **camelCase:** PASS — `_INTEREST_CAP_BALANCE` is a module-level constant (matching the existing `INTEREST_RATE`/`MAX_INTEREST_PER_DAY` ALL_CAPS convention in `timeService.py`), and all new methods/locals follow camelCase.

Summary: three small, independent bug fixes (bank negative-amount exploit, stale crowded-crew dialogue, stale interest-rate dialogue), each with a regression test proven to fail before the fix and pass after.

---

PR #171, opened 2026-08-06
Keep a damaged save slot visible, and stop it being overwritten

### LINKED ISSUE #150: A save whose player.json is corrupt disappears from the menu and its slot is offered as "Create New Save"

### LINKED ISSUE #170: An empty player.json loads as a starting character on the saved calendar, with no warning
## Summary

During a triage pass, a zero-byte `player.json` was found to skip the load machinery entirely rather than failing it. The slot is still offered as loadable, the player is silently handed a starting character, and the empty file is overwritten on the first action.

The guard is a size check, not a load attempt (`src/fishE.py:65`):

```python
player_path = self.saveFileManager.get_save_path("player.json")
if os.path.exists(player_path) and os.path.getsize(player_path) > 0:
    self.loadPlayer()
```

An empty file passes `os.path.exists` and fails `getsize(...) > 0`, so `loadPlayer()` is never called, nothing is appended to `failedLoads`, and neither the dialogue nor the `damaged-...` copy added by #169 fires. `stats.json` and `timeService.json` are non-empty and load normally, so the run continues on the saved calendar and the saved lifetime stats with a brand-new player underneath them.

`SaveFileManager._read_save_metadata` has the same size guard at `src/saveFileManager.py:72`, so it returns a metadata dict carrying `day`, `time` and `last_modified` but no `money`, `fishCount` or `energy`. That dict is truthy, so the slot is listed.

Reproduced against the real `SaveFileManager` and the real `FishE.__init__`, with a slot holding an empty `player.json` beside an intact `stats.json` (900 fish caught, $8,000 earned) and `timeService.json` (day 40):

```
list_save_files() -> [{'slot': 1, ..., 'metadata': {'day': 40, 'time': 9, 'last_modified': ...}}]
failedLoads -> []
showDialogue called -> 0
damaged backups -> []
player.money -> 20   player.fishCount -> 0
stats.totalFishCaught -> 900
timeService.day -> 40
player.json after one save -> {"fishCount": 0, ..., "money": 20, ..., "boats": [], "homeTier": 0, ...}
```

The menu therefore renders `Load Slot 1 (Day 40, $0, 0 fish)` — because `_selectSaveFile` falls back to `metadata.get("money", 0)` and `metadata.get("fishCount", 0)` (`src/fishE.py:159-167`) — and picking it starts a day-40 character with $20, no boat and no home.

## Why it matters

The state the player is left in is incoherent rather than merely reset. It is day 40 with 900 lifetime fish caught and $8,000 lifetime earnings on the stats screen, held by someone with $20, no boat and nowhere to sleep. `progression.catchUp` then runs over the loaded stats, so features earned on lifetime numbers stay open around that character — `unlockedFeatures` comes out as `['shop', 'home', 'housing', 'tavern']` — while everything gated on what the player actually owns is gone. It reads as a save that has been robbed rather than one that failed.

Nothing is said at any point, and the empty file is replaced by the fresh player on the first `save()`, so the only evidence that a run existed is a `stats.json` nobody is prompted to look at.

This is distinct from the two issues nearby:

- #150 is a *corrupt but non-empty* `player.json`, which makes the slot disappear from the menu. Here the slot is listed and picked.
- #143 (closed by #169) is a load that is *attempted and fails*. Here no load is attempted, so the dialogue and the `damaged-...` copy #169 added are both bypassed.

An empty file is a likely artefact of the old truncating write, which #169 replaced — so existing damaged slots on disk can still be in this state even though new ones will not be created this way.

## Suggested fix

Treat an empty save file as a failed load rather than as an absent one. Dropping the `getsize(...) > 0` half of each of the three guards at `src/fishE.py:65`, `:69` and `:73` would send an empty file into `loadPlayer()`, where `json.load` raises `JSONDecodeError`, which is already caught — so it would be recorded in `failedLoads`, reported through the front-end, and the slot copied aside, with no new machinery needed.

Worth checking alongside: whether a slot whose `player.json` carries no usable metadata should be listed as loadable at all (`src/saveFileManager.py:72`), which overlaps with #150's suggested marker.

_Filed by Claude during a Gardener session (https://github.com/Stephenson-Software/gardener); claims above were verified by running the code shown._

dmccoystephenson:
## Summary

Two ways a damaged save could cost the player their run, both of which ended with the damaged slot being overwritten rather than reported.

**A `player.json` that would not parse (#150)** made `_read_save_metadata` return `None`. That dropped the slot from `list_save_files()`, and `get_next_available_slot()` derives the taken slot numbers from that same filtered list — so the slot vanished from the menu *and* was handed straight back as `Create New Save`. Picking the only option offered pointed `select_save_slot` at the occupied directory, and the next `save()` destroyed the intact `stats.json` and `timeService.json` sitting beside the damaged file.

**An empty `player.json` (#170)** failed differently: the `os.path.getsize(...) > 0` half of each load guard skipped the read entirely, so `loadPlayer()` never ran, nothing reached `failedLoads`, and neither the warning dialogue nor the `damaged-...` copy added by #169 fired. The player was handed a starting character on the saved calendar and the empty file was overwritten on the first action.

## What changed

- `_read_save_metadata` returns an `{"unreadable": True, ...}` marker instead of `None`, so a damaged slot stays listed **and** stays claimed. Those two halves have to move together: visibility alone would not stop the overwrite, and counting it alone would not tell the player it exists.
- `player.json` is now read **strictly** — no size guard, since an empty file is damaged rather than absent. Valid JSON that is not an object is also treated as unreadable, because `player_data.get(...)` would raise `AttributeError`, which is not among the caught errors and would take the whole save menu down over one bad slot.
- `timeService.json` is read **tolerantly**. A damaged calendar is not a damaged run: the player still loads, and `FishE` already reports and preserves that damage on load, so the slot stays loadable and the label just falls back to its `Day 1` default.
- The save menu lists the slot as `Slot N - damaged, cannot be loaded` and marks it unpickable with a reason, reusing the `unavailableOptions` mechanism from #168. Deleting is how the slot is reclaimed, so the reason says so and the delete menu tags that row `(damaged)`.
- The three load guards test existence only, so an empty file reaches the loader, raises `JSONDecodeError`, and is reported and copied aside by the machinery #169 already added — no new reporting path.

The two mechanisms stay complementary rather than redundant: the menu catches what it can see cheaply (unparseable `player.json`), while the load path still covers damage the metadata read cannot detect — a save that parses but fails schema validation, or a bad `stats.json`/`timeService.json`.

## Verification

Reproduced both issues' own repro scenarios against the fixed code. Issue #150's slot (truncated `player.json` beside intact `stats.json`/`timeService.json`):

```
list_save_files() -> [{'slot': 1, ..., 'metadata': {'unreadable': True, 'reason': 'Unterminated string ...'}}]
get_next_available_slot() -> 2
slot_1 still on disk -> ['player.json', 'stats.json', 'timeService.json']
menu options -> ['Slot 1 - damaged, cannot be loaded', 'Create New Save (Slot 2)', 'Delete a Save File', 'Quit']
unavailable  -> {1: 'this save is damaged - delete it to reuse the slot'}
slot selected -> 2
```

Issue #170's slot (empty `player.json`):

```
failedLoads -> ['player.json (Expecting value: line 1 column 1 (char 0))']
showDialogue called -> 1
damaged backups -> ['damaged-20260806-071420']
backup contents -> ['player.json', 'stats.json', 'timeService.json']
```

## Test plan

- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest --verbose -vv --cov=src --cov-report=term-missing --cov-report=xml:cov.xml` — **791 passed**, 97% total coverage; every new branch covered
- [x] Two existing tests that asserted the old behaviour (`test_read_save_metadata_corrupted_json`, `test_read_save_metadata_empty_player_file`) were updated to the marker — that behaviour was the bug
- [x] **Front-end parity:** the change is at the shared `showOptions(descriptor, options, unavailable)` call site, and all three implementations (`UserInterface`, `PygameUserInterface`, `WebUserInterface` — with `ConsoleUserInterface`/`PyodideUserInterface` inheriting) already honour `unavailableOptions` from #168; `web/client.js` already greys the row and shows the reason. No front-end-specific code was needed or added.
- [x] **Schemas:** no `Player`/`Stats`/`TimeService` field was added, renamed or retyped, so no `schemas/*.json` change applies; cross-checked all three schemas against their `*JsonReaderWriter` anyway and they are in sync.
- [x] `README.md`'s save-file section updated to match (it described the pre-#150 menu); `PLANNING.md` needed no change.
- [x] `black`/`autoflake` run per `format.sh`. Note: this sandbox's `black` is 23.1.0 and reformats four files this PR does not touch (`src/business/export.py`, `src/location/shop.py`, `src/npc/villagers.py`, `tests/ui/test_consoleUserInterface.py`) — that churn was reverted and left out. The five files here are `black --check` clean.

Closes #150
Closes #170

---


src/saveFileManager.py
```diff
@@ -58,41 +58,87 @@ def list_save_files(self):
...
-
-            return metadata
-        except (json.JSONDecodeError, IOError, OSError) as e:
-            # Return None for corrupted or inaccessible save files
+            with open(time_file, "r") as f:
+                time_data = json.load(f)
+            if isinstance(time_data, dict):
+                metadata["day"] = time_data.get("day", 1)
+                metadata["time"] = time_data.get("time", 0)
+        except (json.JSONDecodeError, IOError, OSError):
```
dmccoystephenson, line 108: Deliberate asymmetry, flagged so it doesn't read as a copy-paste slip against the strict `player.json` read 20 lines above: this one swallows the error and keeps the slot loadable.

`timeService.json` is not the run. A slot whose calendar is damaged still holds a readable player, and `FishE.loadTimeService` already records the failure in `failedLoads`, reports it through the front-end and copies the slot aside. Marking the slot unreadable here would take a mostly-intact run and lock it out of the menu entirely — strictly worse for the player than losing the day counter.

The cost is that `day`/`time` are simply absent, which `_selectSaveFile` renders with its own `Day 1` default, so the label can understate a damaged-calendar slot. That seemed better than refusing to load it.

src/saveFileManager.py
```diff
@@ -58,41 +58,87 @@ def list_save_files(self):
...
+        # Deliberately no "size > 0" guard: an empty file is a damaged file
+        # rather than an absent one, and skipping the read for it is what let a
+        # zero-byte save be offered in the menu as a real one.
+        try:
+            with open(player_file, "r") as f:
+                player_data = json.load(f)
+        except (json.JSONDecodeError, IOError, OSError) as error:
+            return self._unreadable_save_metadata(player_file, error)
+
+        if not isinstance(player_data, dict):
```
dmccoystephenson, line 85: This guard is not defensive padding — it closes a real crash that the original `except` clause could not catch.

`json.load` on a `player.json` holding valid JSON that isn't an object (`42`, `null`, `"x"`, or some unrelated file copied over the save) returns a non-dict, and the `player_data.get(...)` calls below raise `AttributeError`. That is not in `(json.JSONDecodeError, IOError, OSError)`, so it would propagate out of `_read_save_metadata` and out of `list_save_files` — taking the entire save menu down on startup over one bad slot, which is a strictly worse failure than the disappearing slot #150 is about.

Covered by `test_read_save_metadata_player_file_that_is_not_an_object`, which asserts the listing still works rather than only checking the return value.

src/fishE.py
```diff
@@ -190,10 +219,30 @@ def _selectSaveFile(self):
                 # loop to show the refreshed menu either way
             elif kind == "quit":
                 exit(0)
+            elif kind == "damaged":
```
dmccoystephenson, line 222: This branch should be unreachable and is here on purpose; worth justifying rather than leaving as apparently-dead code.

`showOptions` is contracted not to return an unavailable option's number, and `unavailableReasons`' all-unavailable fallback cannot fire here because `Quit` is always selectable. But without a branch, `kind == "damaged"` falls out of the `if`/`elif` chain and the `while True` simply re-renders the same menu — the player picks the row, and the game appears to ignore them, with nothing said. That is an unexplained hang rather than a visible bug.

I hit exactly this while writing the reproduction script for #150 (a mock returning `"1"` unconditionally span the menu forever), which is what convinced me not to rely on the contract. A new front-end is the realistic way this gets violated for real, and front-end parity is the most common gap in this repo. Covered by `test_selectSaveFile_explains_a_damaged_slot_a_front_end_let_through`.

src/saveFileManager.py
```diff
@@ -58,41 +58,87 @@ def list_save_files(self):
...
+            # file copied over the save - has no fields to read, and .get()
+            # would raise AttributeError here, taking the whole menu down with
+            # it rather than reporting one bad slot.
+            return self._unreadable_save_metadata(
+                player_file, ValueError("not a JSON object")
+            )
+
+        metadata["money"] = player_data.get("money", 0)
+        metadata["fishCount"] = player_data.get("fishCount", 0)
+        metadata["energy"] = player_data.get("energy", 100)
```
dmccoystephenson, line 96: Observation, pre-existing and not changed by this PR (the line moved but the behaviour is untouched): `metadata["energy"]` has no production consumer. `_selectSaveFile` reads only `day`, `money` and `fishCount`, and the only other reference in the repo is an assertion in `test_read_save_metadata_partial_data`.

Left alone deliberately — removing it is unrelated to #150/#170 and would widen this diff. Worth a small follow-up issue to either drop it or put it on the slot label, where it would arguably be more useful than the fish count.

dmccoystephenson (commented):
Self-review: read the full diff against the two issues and the surrounding save/load machinery. No correctness problems found; the notes below record the three judgement calls a later reader is most likely to want to undo, plus one pre-existing observation worth a follow-up issue rather than a change here.

One finding from this pass was already fixed in d8b0b12 rather than left as a comment: the damaged row put the blocker in *both* the label and the reason, so a front-end that appends the reason rendered `Slot 1 - damaged, cannot be loaded (unavailable: this save is damaged - delete it to reuse the slot)`. The label now only identifies the slot, matching how every other unusable option in the game is built.

Deliberately **not** changed: the `_read_save_metadata` marker is not derived from schema validation. Keeping the metadata read cheap and limited to "does this parse" leaves a schema-invalid save loadable, which is correct — #169 already reports and preserves that case on load, and a save whose values are out of range is far more recoverable than one that is not JSON. Coupling `SaveFileManager` to the reader/writers and `schemas/` to catch it in the menu would be a worse trade.

PR #172, opened 2026-08-07
Report the overnight fleet news everywhere a day rolls over

### LINKED ISSUE #151: The overnight fleet report is discarded at three of the six places a day rolls over

dmccoystephenson:
## Summary

`TimeService.increaseDay()` returns `{"evicted": bool, "report": [str]}`, and its own docstring states the contract: *"a pirate crew can come home a man short, the player has to be told rather than left to notice their roster changed."* `README.md:75` makes the same promise. Three of the six places a day rolls over honoured it; three read only `evicted` and dropped `report`.

**Fixed — the three that dropped it:**

- **A captained voyage leg** (`docks.py`) bound *nothing* from `increaseDay()`. Each leg runs wages, rent, crew walkouts and possible eviction, so a player could sail a five-day voyage and come home homeless without a single line of text saying so.
- **A fishing trip long enough to cross 8am** read only `evicted`, so a crew member who didn't come back went unmentioned until the player happened to open Manage Fleet.
- **`FishE`'s hourly tick** — the one place guaranteed to run whatever the player did — read only `evicted`.

## What changed

The three sites that *did* work were three hand-rolled copies of the same loop (`home.py` and `tavern.py` were character-identical to `Docks._reportTheDay`'s body). All six now go through two functions added next to the producer in `timeService.py`:

- `dayReportLines(summary)` — the display lines: the fleet's takings, then the eviction notice. Lives beside `increaseDay` so reading one half of the contract and dropping the other stops being the easy thing to write. Tolerates a summary with no `report` key, so callers (and existing tests) that only ever produced an eviction flag don't raise.
- `appendDayReport(prompt, summary, separator=" ")` — appends those lines to a prompt.

`Docks._reportTheDay` — a fourth copy, then a one-line passthrough — is gone; its single call site calls the shared helper directly.

Two call sites needed more than a swap:

- **The voyage leg** now rolls the day *before* building its dialogue, so the news lands on the same screen as the leg that caused it rather than a screen later. The `"\n".join([outcome] + notes) if notes else outcome` conditional was redundant (joining a one-item list returns the item) and collapsed into the unconditional join.
- **The fishing trip** collects its lines across the hour loop and folds them into the trip report, because that report is written *after* the loop and would otherwise overwrite anything put on the prompt during it. A trip is 1–10 hours, so it crosses at most one 8am boundary and the lines can't stack up.

`separator` exists because `FishE`'s loop deliberately uses a wider gap — there the day's news is appended to text a *different* subsystem already wrote, which isn't true of the location sites. This also resolves the spacing inconsistency the issue flagged, by making it an explicit parameter with a stated reason rather than an accident.

## Verification

The existing 791 tests all passed **before** this fix, which is why the bug survived — no test covered the reporting at any of the six sites except the eviction flag. So the new tests were checked against a simulated regression: with `dayReportLines` stubbed to drop the report, exactly these fail, and nothing else does.

```
test_fish_reports_what_the_fleet_did_while_the_player_was_on_the_water
test_a_voyage_leg_reports_what_happened_back_home
test_sleep_reports_what_the_fleet_did_overnight
test_getDrunk_reports_what_the_fleet_did_overnight
test_play_appends_the_overnight_fleet_report
test_play_appends_the_fleet_report_and_the_eviction_together
test_dayReportLines_carries_the_fleet_report_and_the_eviction_notice
test_appendDayReport_separator_defaults_to_one_space
test_appendDayReport_honours_a_wider_separator
```

All six sites are now guarded: the export site's existing eviction test was extended to assert the fleet report too, so a future per-site drift fails the build instead of going quiet.

## Test plan

- [x] `SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=dummy python3 -m pytest --verbose -vv --cov=src --cov-report=term-missing --cov-report=xml:cov.xml` — **803 passed**, 97% total coverage
- [x] Confirmed the new tests fail against a simulated re-introduction of the bug (above), so they test the fix rather than the code
- [x] **Front-end parity:** no front-end code touched. The day report reaches the player as prompt text or, on a voyage leg, inside an existing `showDialogue` call — both already front-end agnostic, so console, pygame, web and Pyodide all get it through the same path with no per-UI work.
- [x] **Schemas:** no `Player`/`Stats`/`TimeService` field added, renamed or retyped, so no `schemas/*.json` change applies.
- [x] **Docs:** `README.md:75` already promised this overnight report — the fix makes that promise true at all six sites instead of three, so no wording change was needed. `PLANNING.md` unaffected.
- [x] `black`/`autoflake` per `format.sh`; `housing` became an unused import in `docks.py`, `tavern.py` and `fishE.py` and was removed. Only the ten files in this diff are touched.

Closes #151

---

### DIFF
--- a/src/location/docks.py
+++ b/src/location/docks.py
@@ -3,7 +3,7 @@ import random
 from location.enum.locationType import LocationType
 from player.player import Player
 from prompt.prompt import Prompt
-from world.timeService import TimeService
+from world.timeService import TimeService, appendDayReport, dayReportLines
 from stats.stats import Stats
 from ui.userInterface import UserInterface
 from npc.npc import NPC
@@ -13,7 +13,6 @@ from business import business
 from business import boats
 from business import export
 from business import adventures
-from housing import housing
 from progression import progression
 
 
@@ -920,11 +919,15 @@ class Docks:
             outcome = adventures.resolveChoice(voyage, choices[picked - 1])
             notes = adventures.advanceLeg(voyage)
 
-            self.userInterface.showDialogue(
-                "\n".join([outcome] + notes) if notes else outcome
-            )
-            # A day at sea is a day at home too.
-            self.timeService.increaseDay()
+            # A day at sea is a day at home too - the fleet still lands its
+            # catch and rent still comes due while the player is away. The
+            # day rolls before the dialogue is built so its news goes on the
+            # same screen as the leg that caused it; dropping it is how a
+            # player could sail a five-day voyage and come home evicted with
+            # nothing on screen having said so.
+            dayLines = dayReportLines(self.timeService.increaseDay())
+
+            self.userInterface.showDialogue("\n".join([outcome] + notes + dayLines))
 
         self._endVoyage(voyage)
 
@@ -972,14 +975,6 @@ class Docks:
             else "  Hull     sound"
         )
         return "\n".join(lines)
-
-    def _reportTheDay(self, summary):
-        """Append what happened while a day passed - the fleet's overnight
-        takings, and anything that went wrong at home."""
-        for line in summary.get("report", []):
-            self.currentPrompt.text += " " + line
-        if summary["evicted"]:
-            self.currentPrompt.text += " " + housing.EVICTION_MESSAGE
 
     def _exportStatus(self):
         """What the next export run would look like, shown above the market
@@ -1076,7 +1071,7 @@ class Docks:
         # The round trip is what stops exporting from being a free repeatable
         # action, so the day passes here (and everything a new day brings -
         # the crew's catch, wages, rent - happens while you're away).
-        self._reportTheDay(self.timeService.increaseDay())
+        appendDayReport(self.currentPrompt, self.timeService.increaseDay())
         return True
 
     def _exportRefusal(self, summary, market):
@@ -1181,14 +1176,17 @@ class Docks:
             quality, qualityLabel = 0.25, "The fish nearly got away."
 
         # Spend the fishing hours: time passes and energy is consumed. A long
-        # enough trip can cross a day boundary (and so, e.g., miss a rent
-        # payment) without the player ever seeing a "new day" screen, so
-        # track that across the loop to mention it in the trip's own report.
-        evicted = False
+        # enough trip can cross a day boundary - missing a rent payment, or
+        # landing the fleet's catch - without the player ever seeing a "new
+        # day" screen, so the day's news is collected here and folded into the
+        # trip's own report below. It has to be collected rather than appended
+        # as it happens, because that report is written further down and would
+        # overwrite anything put on the prompt now. A trip runs 1-10 hours, so
+        # it crosses at most one 8am boundary and these lines can't stack up.
+        dayLines = []
         for i in range(hours):
             self.stats.hoursSpentFishing += 1
-            if self.timeService.increaseTime()["evicted"]:
-                evicted = True
+            dayLines += dayReportLines(self.timeService.increaseTime())
             self.player.spendEnergy(10)  # Consume 10 energy per hour
 
         baseFish = random.randint(1, 10)
@@ -1216,8 +1214,10 @@ class Docks:
         if weatherLabel:
             self.currentPrompt.text += " " + weatherLabel
 
-        if evicted:
-            self.currentPrompt.text += " " + housing.EVICTION_MESSAGE
+        # What the fleet did while the player was out on the water, and the
+        # eviction notice if a day turned over mid-trip.
+        for line in dayLines:
+            self.currentPrompt.text += " " + line
 
     def talkToNPC(self):
         self.userInterface.showInteractiveDialogue(self.npc)
--- a/src/world/timeService.py
+++ b/src/world/timeService.py
@@ -73,3 +73,35 @@ class TimeService:
             "evicted": rentSummary["evicted"],
             "report": boats.describeDay(fleetSummary),
         }
+
+
+def dayReportLines(summary):
+    """What the player has to be told about a day that just passed.
+
+    Takes increaseDay()'s (or increaseTime()'s) return value and turns it into
+    display lines: the fleet's overnight takings, then the eviction notice if
+    rent went unpaid.
+
+    Lives beside the producer because reading only "evicted" and dropping
+    "report" is the mistake this is here to stop being possible - a player
+    could sail a multi-day voyage and come home a crew member short, or
+    homeless, without a line of text saying so. increaseTime() returns
+    {"evicted": False, "report": []} on the hours that don't roll a day, so
+    every caller can pass its return value straight in without first checking
+    whether a day actually turned over; the result is simply empty."""
+    lines = list(summary.get("report", []))
+    if summary.get("evicted"):
+        lines.append(housing.EVICTION_MESSAGE)
+    return lines
+
+
+def appendDayReport(prompt, summary, separator=" "):
+    """Append dayReportLines() to the prompt the player is about to be shown.
+
+    separator is the gap placed before each line. It defaults to one space,
+    which is what a location uses when the line continues a sentence it just
+    wrote ("You sleep until the next morning. ..."). FishE's game loop passes
+    two, because there the day's news is being appended to whatever a
+    completely different subsystem already put on the screen."""
+    for line in dayReportLines(summary):
+        prompt.text += separator + line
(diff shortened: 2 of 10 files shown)


src/location/docks.py
```diff
@@ -920,11 +919,15 @@ def _sailVoyage(self, voyage):
...
-            )
-            # A day at sea is a day at home too.
-            self.timeService.increaseDay()
+            # A day at sea is a day at home too - the fleet still lands its
+            # catch and rent still comes due while the player is away. The
+            # day rolls before the dialogue is built so its news goes on the
+            # same screen as the leg that caused it; dropping it is how a
+            # player could sail a five-day voyage and come home evicted with
+            # nothing on screen having said so.
+            dayLines = dayReportLines(self.timeService.increaseDay())
```
dmccoystephenson, line 928: This is the one site where the fix reorders existing calls rather than just reading more of a return value, so it deserves the closest look.

`increaseDay()` used to run *after* `showDialogue`, so the day's news could not have gone on that screen even if it had been bound. Rolling the day first is what lets it land on the same screen as the leg that caused it, instead of a screen later or not at all.

What I checked: nothing between the old and new position reads the clock or the player's money, so moving it changes only *when the text appears*, not what the leg resolves to. `adventures.advanceLeg` is still called before the roll, and `adventures.isOver(voyage)` is driven by legs sailed rather than by `timeService.day`, so the loop's termination is unaffected — `test_a_voyage_sails_every_leg_costs_days_and_comes_home` still asserts one day per leg.

The `"\n".join([outcome] + notes) if notes else outcome` conditional was also dropped: joining a one-item list returns the item, so the ternary was always equivalent to the join.

src/location/docks.py
```diff
@@ -1181,14 +1176,17 @@ def fish(self):
...
-        # track that across the loop to mention it in the trip's own report.
-        evicted = False
+        # enough trip can cross a day boundary - missing a rent payment, or
+        # landing the fleet's catch - without the player ever seeing a "new
+        # day" screen, so the day's news is collected here and folded into the
+        # trip's own report below. It has to be collected rather than appended
+        # as it happens, because that report is written further down and would
+        # overwrite anything put on the prompt now. A trip runs 1-10 hours, so
+        # it crosses at most one 8am boundary and these lines can't stack up.
+        dayLines = []
```
dmccoystephenson, line 1186: The one site that can't use `appendDayReport` directly, flagged so it isn't "simplified" into a call that silently loses the report.

The trip's own summary is assigned with `=` further down (`self.currentPrompt.text = "You caught %d ..."`), not appended — so anything written to the prompt inside this loop is overwritten a few lines later. The lines have to be collected and appended *after* that assignment, which is why this reads `dayLines +=` here and a plain loop at the end rather than the shared helper.

The previous `evicted = True` flag collapsed repeats implicitly; a list doesn't. That's safe because `hours = random.randint(1, 10)`, so a trip crosses at most one 8am boundary — but it is the assumption to re-check if the hour range ever widens.

src/world/timeService.py
```diff
@@ -73,3 +73,35 @@ def increaseDay(self):
...
+    {"evicted": False, "report": []} on the hours that don't roll a day, so
+    every caller can pass its return value straight in without first checking
+    whether a day actually turned over; the result is simply empty."""
+    lines = list(summary.get("report", []))
+    if summary.get("evicted"):
+        lines.append(housing.EVICTION_MESSAGE)
+    return lines
+
+
+def appendDayReport(prompt, summary, separator=" "):
```
dmccoystephenson, line 98: `separator` is a parameter used by exactly one caller, which normally isn't worth the knob — so, the reasoning:

#151 flagged that `fishE.py` appended with two spaces where the other five sites use one, reading like an inconsistency to be normalised away. It isn't. In a location the line continues a sentence that method just wrote ("You sleep until the next morning. The fleet landed..."), whereas in `FishE`'s game loop it is appended to whatever a *different* subsystem already put on the screen, alongside the milestone and unlock announcements that use the same wider gap.

So rather than flattening both to one space (which would visually merge two unrelated messages) or leaving five copies of a magic `" "`, the difference is now a named parameter with the reason attached. Both spacings are asserted — `test_appendDayReport_separator_defaults_to_one_space` and `..._honours_a_wider_separator`.

src/world/timeService.py
```diff
@@ -73,3 +73,35 @@ def increaseDay(self):
             "evicted": rentSummary["evicted"],
             "report": boats.describeDay(fleetSummary),
         }
+
+
+def dayReportLines(summary):
```
dmccoystephenson, line 78: Two scope notes on where this landed.

**Why `timeService` and not a UI module:** these are module functions on the same file as the `increaseDay()` whose return value they interpret. That adjacency is the actual fix — the bug was six callers each deciding how to read a two-key dict, and half of them reading one key. `timeService.py` already imports `housing`, so `EVICTION_MESSAGE` costs no new dependency. They take a duck-typed `prompt` (anything with `.text`) rather than importing `Prompt`, so this stays free of front-end coupling.

**Why `.get("report", [])` rather than requiring the key:** existing tests mock `increaseTime` as `{"evicted": False}` with no `report` key at all (e.g. `test_fish_mentions_eviction_when_a_day_rolls_over_mid_trip`). Being strict here would have failed those tests for reasons unrelated to what they assert, and would make the helper hostile to any caller that hasn't been updated yet. `test_dayReportLines_tolerates_a_summary_with_no_report_key` pins that tolerance so it isn't tightened by accident.

dmccoystephenson (commented):
Self-review: read the full diff against #151 and traced each of the six rollover sites. No correctness problems found. Notes below record the two places where the fix is not a straight swap (and so is the most likely thing a later reader breaks), one deliberate behaviour change worth calling out explicitly, and one scope decision.

The thing I'd want a reviewer to check hardest is the ordering change on the voyage leg — it is the only site where this PR moves an existing call rather than just reading more of its return value.

Worth stating plainly, since the diff doesn't show it: **all 791 pre-existing tests passed before this fix**. Nothing covered the report at any of the six sites, only the eviction flag — which is exactly how three sites drifted. So I checked the new tests against a simulated re-introduction of the bug (`dayReportLines` stubbed to drop the report) and confirmed the nine expected tests fail and nothing else does. Without that step these tests would assert current behaviour rather than the fix.

PR #175, opened 2026-08-09
Release the front-end wherever the game ends, and let the browser see the ended screen

### LINKED ISSUE #153: The server-backed web front-end shows "lost connection" instead of the ended screen when the game finishes

dmccoystephenson:
## Summary

- `cleanup()` is now called once, from a `finally` in `FishE.play()`, rather than from the Pyodide entry point alone. Every way a run can finish — retiring, quitting, an unhandled error — is covered, for all four front-ends. It stays a no-op for the console.
- `exit(0)` in the save-file menu's "Quit" branch has been replaced by clearing `running`. Ending the interpreter from inside `__init__` meant no front-end was ever released; `__init__` now returns early and `play()` does nothing but clean up.
- The server-backed web front-end held the browser's ended screen for as long as it took to publish it and no longer: the socket was closed before the next 300&nbsp;ms poll could fetch it, so a player who had just retired was shown "Lost connection to the game". The server is now kept open until that screen has actually been sent (typically one poll interval), with a two-second timeout for when nothing is listening at all.
- The page stops polling once the ended screen is rendered, instead of failing against a finished process for the rest of the tab's life.
- `web/pyodide_main.py`'s `SystemExit` handling and its own `cleanup()` call have been dropped as redundant.

## Test plan

- [x] `python3 -m compileall -q src tests web examples`
- [x] `python3 -m pytest --cov=src --cov-report=term-missing --cov-report=xml:cov.xml` — 815 passed, `src/ui/webUserInterface.py` at 97%
- [x] Front-ends checked: the change is in shared code (`FishE.play`), so console, pygame, `UIType.WEB` and `UIType.PYODIDE` all take the same path. Console and pygame `cleanup()` were read and are unaffected beyond now being called (pygame's `pygame.quit()` is idempotent, and its own QUIT handlers still call it). The delivery wait is skipped when no server is running, which is how the Pyodide subclass inherits it.
- [x] Nine tests added: `play()` cleans up on a normal end and on an exception; "Quit" from the save-file menu ends the run without `SystemExit` and claims no slot; the ended screen is published with or without a server; the server is held open until that screen is fetched and gives up when nothing fetches it; delivery bookkeeping ignores a late overlapping poll; the client's poll loop stops on `ended`.
- [x] Documentation checked against the implementation: `README.md`, `PLANNING.md` and the three `schemas/*.json` files make no claim touched by this change (no `Player`/`Stats`/`TimeService` field was added, renamed or retyped).

Closes #153

### DIFF
--- a/src/fishE.py
+++ b/src/fishE.py
@@ -59,6 +59,14 @@ class FishE:
         # Show save file selection menu (uses the UI above)
         self._selectSaveFile()
 
+        # "Quit" from that menu clears running rather than ending the process,
+        # so the front-end still gets its cleanup() (see _selectSaveFile and
+        # play()). Nothing below can run without a chosen slot - get_save_path()
+        # raises when none was selected - and there is no run to build anyway,
+        # so construction stops here and play() returns immediately.
+        if not self.running:
+            return
+
         # Load the chosen slot over the defaults if it has data.
         #
         # Existence is the only condition: a file that is present but empty is a
@@ -153,7 +161,10 @@ class FishE:
         """Display the save-file menu through the UI and let the player choose.
 
         Slots and actions are presented as numbered options (so the menu renders
-        and reads input through the active front-end — console or pygame)."""
+        and reads input through the active front-end — console or pygame).
+
+        Returns once a slot is selected, or with self.running cleared if the
+        player chose "Quit" without picking one."""
         while True:  # loop instead of recursion to avoid stack overflow
             save_files = self.saveFileManager.list_save_files()
 
@@ -217,7 +228,16 @@ class FishE:
                 self._deleteSaveFile(save_files)
                 # loop to show the refreshed menu either way
             elif kind == "quit":
-                exit(0)
+                # Ending the run rather than the process. exit(0) killed the
+                # interpreter from inside __init__, so cleanup() was never
+                # reached: the pygame window vanished without pygame.quit(),
+                # and both browser front-ends left the tab on the save-file
+                # menu with no ended screen (the Pyodide entry point had to
+                # catch SystemExit and post one itself). __init__ returns
+                # early on a cleared running flag, and play() then does
+                # nothing but clean up - one exit path for every front-end.
+                self.running = False
+                return
             elif kind == "damaged":
                 # A conforming front-end refuses to return an unavailable
                 # option's number, so this should be unreachable. It is handled
@@ -266,6 +286,21 @@ class FishE:
         return False
 
     def play(self):
+        """Run the game loop, releasing the front-end however it ends.
+
+        The cleanup() call is here, once, rather than at each of the places a
+        run can finish (retiring, quitting, an unhandled error): every
+        front-end needs it and only this method sees all of those endings.
+        Without it the pygame window closed without pygame.quit(), and the
+        browser front-ends never published their ended screen - the tab kept
+        polling a server that had exited and told a player who had just retired
+        that the connection was lost. It is a no-op for the console."""
+        try:
+            self._runGameLoop()
+        finally:
+            self.userInterface.cleanup()
+
+    def _runGameLoop(self):
         while self.running:
             # show the current location and goal progress in the UI header
             self.userInterface.currentLocationName = self.currentLocation.capitalize()
--- a/src/ui/webUserInterface.py
+++ b/src/ui/webUserInterface.py
@@ -247,7 +247,7 @@ class WebUserInterface(BaseUserInterface):
         with self._lock:
             self._deliveredVersion = max(self._deliveredVersion, version)
 
-    def awaitScreenDelivery(self, timeout):
+    def _awaitScreenDelivery(self, timeout):
         """Block until the current screen has been sent to the browser.
 
         Returns True if it went out, False if the timeout ran out first -
@@ -356,7 +356,7 @@ class WebUserInterface(BaseUserInterface):
             # Hold the server open until the page has the ended screen;
             # otherwise the run's last screen is never fetched and the browser
             # reports a lost connection instead.
-            self.awaitScreenDelivery(self._endedScreenTimeoutSeconds)
+            self._awaitScreenDelivery(self._endedScreenTimeoutSeconds)
             self._server.shutdown()
             self._server.server_close()
             self._server = None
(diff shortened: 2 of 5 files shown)


dmccoystephenson:
## Self-review

The full diff was read against this repo's conventions (front-end parity, the schema/reader-writer contract, test coverage of new public methods, doc drift). One finding was acted on before this comment was posted; the rest are notes on deliberate choices, and nothing is left outstanding.

**Acted on**

- `src/ui/webUserInterface.py:250` — the ended-screen wait was originally added as `awaitScreenDelivery`, which reads as part of this class's public surface. Every other internal helper here carries the underscore (`_present`, `_awaitInput`, `_header`); only `get_state`, `submit_input` and the new `record_state_delivered` are public, because the request handler calls them. It has been renamed to `_awaitScreenDelivery` (commit 27829c9).

**Reviewed and judged correct**

- `src/fishE.py:296` — the loop body was extracted to `_runGameLoop()` rather than re-indented under a `try:`, so the diff shows what changed instead of fifty moved lines. The behaviour is identical, and `play()` remains the only entry point anything calls.
- `src/fishE.py:67` — the early return leaves `self.locations` and `self.currentLocation` unset when "Quit" is chosen before a slot is. That is reachable only through `_runGameLoop()`, which never runs with `running` cleared, and building five locations for a game nobody will play would be the stranger choice. `get_save_path()` raising "No save slot selected" is what makes stopping here mandatory rather than optional.
- `src/ui/pygameUserInterface.py:196` and its four siblings — a window-close during the game now reaches `cleanup()` twice: once from the QUIT handler, then again through `play()`'s `finally` as `SystemExit` unwinds. `pygame.quit()` is documented as safe to call on an uninitialised pygame, so this is a no-op rather than a fault, and the handlers have to stay because a QUIT during `__init__`'s save-file menu never reaches `play()` at all.
- `src/ui/webUserInterface.py:160` — delivery is recorded after `_send()` returns rather than inside `get_state()`. Recording it at snapshot time would let `cleanup()` close the server while the response for that very snapshot was still being written. A client that disconnects mid-write leaves the version unrecorded, which correctly reads as "not delivered".
- `src/ui/webUserInterface.py:246` — `max()` rather than assignment, because two polls can overlap and the older one can finish last; a stale write would otherwise un-deliver the ended screen and cost the full timeout.
- `src/ui/webUserInterface.py:359` — the two-second timeout is only ever paid when nothing is polling (a closed tab, or the game driven by something other than a browser). With a page open the wait ends within one 300&nbsp;ms poll. `makeWebUI` in the tests passes a short timeout for the same reason, so no test sits on the production default.
- `web/pyodide_main.py` — the `SystemExit` branch was removed rather than kept as a safety net, because the only thing that raised it was the `exit(0)` this PR deletes. The end-to-end test that covers quitting from a Worker (`tests/ui/test_pyodideUserInterface.py:369`) still passes, now getting its ended screen from `play()` instead.

**Parity, schemas and docs**

- No front-end gap: the fix lives in `FishE.play()`, which is shared, and the one front-end-specific part (holding the socket open) sits in `WebUserInterface` where `PyodideUserInterface` inherits it as a no-op — it has no server, so `cleanup()` returns straight after publishing the screen.
- No `Player`, `Stats` or `TimeService` field was touched, so `schemas/player.json`, `schemas/stats.json` and `schemas/timeService.json` need no change.
- `README.md` and `PLANNING.md` were re-read against the implementation; neither describes what happens when a run ends, so nothing there became inaccurate.

This review was performed during a Gardener session (https://github.com/Stephenson-Software/gardener).

PR #177, opened 2026-08-12
Name FISHE_WEB_PORT when a web port is busy or misspelled

### LINKED ISSUE #176: A busy or misspelled port is reported as a raw traceback by both web entry points
## Summary

`FISHE_WEB_PORT` is read in three places, and only one of them explains itself when the value cannot be used.

- `src/ui/userInterfaceFactory.py:44-50` — the model. A non-numeric value is caught and named:
  ```python
  port_str = os.environ.get("FISHE_WEB_PORT", "8000")
  try:
      port = int(port_str)
  except ValueError:
      raise ValueError(f"FISHE_WEB_PORT must be an integer, got: {port_str!r}")
  ```
  Covered by `tests/ui/test_userInterfaceFactory.py:117-125`.

- `web/serve.py:87` — the same variable, converted with no guard at all:
  ```python
  port = int(os.environ.get("FISHE_WEB_PORT", "8080"))
  ```
  `FISHE_WEB_PORT=80801x python3 web/serve.py` therefore dies on `ValueError: invalid literal for int() with base 10: '80801x'`, which names neither the variable nor what it should hold. This is the entry point the `Dockerfile` uses (`ENV FISHE_WEB_PORT=8080`, `Dockerfile:35`), so it is the one most likely to be handed a value from outside.

- Neither entry point handles a port that is simply taken. `web/serve.py:90` (`_Server((host, port), _Handler).serve_forever()`) and `src/ui/webUserInterface.py:222` (`ThreadingHTTPServer((host, port), _makeRequestHandler(self))`) both bind in a constructor with nothing around them, so a second copy of the game — or anything else already on 8000/8080 — surfaces as `OSError: [Errno 98] Address already in use` and a stack trace through `http.server`. `allow_reuse_address` (`web/serve.py:81`) only covers a socket in `TIME_WAIT`; a live listener still fails.

Both failures happen before the front-end exists, so there is no `showDialogue` to say it through — the message has to be printed or raised, the way the factory already does.

## Why it matters

The two failures are the ordinary ones. "I already have it running in another terminal" and "I typed the port wrong" are what a player hits, and in both cases the traceback points at `http.server` or at `int()` rather than at the one environment variable they need to change.

## Suggested fix

Give `web/serve.py` the factory's integer check (or share one helper between them — both read the same variable with different defaults, 8000 and 8080, which is itself worth a comment), and turn a failed bind into a sentence naming the address and `FISHE_WEB_PORT` in both entry points. Tests for both, alongside the existing `tests/ui/test_userInterfaceFactory.py:117` and `tests/web/test_serve.py`.

_Filed by Claude during an automated triage pass; every claim above was verified against source. This issue was filed during a Gardener session (https://github.com/Stephenson-Software/gardener)._

dmccoystephenson:
## Summary

- `web/serve.py` — the entry point the `Dockerfile` runs, and therefore the one most likely to be handed a value from outside — converted `FISHE_WEB_PORT` with a bare `int()`. A typo was reported as `ValueError: invalid literal for int() with base 10: '80801x'`, which names neither the variable nor what it should hold. The integer check `UserInterfaceFactory`'s WEB branch already had has been given to it as well.
- Neither web entry point handled a port that is simply taken. Both bound inside a constructor with nothing around it, so a second copy of the game — or anything else already on 8000/8080 — was surfaced as `OSError: [Errno 98] Address already in use` traced through `http.server`. `allow_reuse_address` does not cover this; only a socket left in `TIME_WAIT` is reopened by it, while a live listener still refuses the bind. A refused bind is now turned into a sentence naming the address, the reason (already listening / not permitted), and `FISHE_WEB_PORT`.
- Both failures happen before the front-end exists, so no `showDialogue` is available to say them through — they are raised, the way the factory already does.
- `web/serve.py` now binds before the URL is announced, so a failure is no longer preceded by an address that was never served.
- The two defaults are kept apart (8080 for the static server, 8000 for the server-backed front-end behind `examples/web_app.py`), since those are separate programs that can be run at the same time; a comment now records why. Only the check on the value is shared in spirit.
- `README.md` now mentions `FISHE_WEB_PORT`/`FISHE_WEB_HOST`, which the new messages instruct the player to set.

## Front-ends touched

Only the two web entry points bind a socket, so this failure path has no console, pygame, or Pyodide equivalent: the console and pygame front-ends open no server, and the Pyodide front-end runs inside the browser tab. No `BaseUserInterface` primitive was changed, so front-end parity is unaffected.

## Test plan

- [x] `python3 -m compileall -q src tests web`
- [x] `python3 -m pytest` — 820 passed, including four new tests in `tests/web/test_serve.py` (default port, port from the environment, a misspelled port naming the variable, a taken port explained rather than traced) and one in `tests/ui/test_webUserInterface.py` (a taken port explained rather than traced)
- [x] `black` run over the changed files
- [x] No `Player`/`Stats`/`TimeService` field changed, so `schemas/*.json` and the `*JsonReaderWriter` classes are untouched

Closes #176

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ python3 web/serve.py
 # then open http://127.0.0.1:8080
 ```
 
+Set `FISHE_WEB_PORT` to serve on a different port (and `FISHE_WEB_HOST`, e.g. `0.0.0.0`, to be reachable from outside the machine); both entry points name that variable if the port they were given is misspelled or already taken.
+
 `web/serve.py` only serves files — it never runs the game. It does have to send the `Cross-Origin-Opener-Policy` and `Cross-Origin-Embedder-Policy` headers it sets, though: without them the page is not cross-origin isolated, `SharedArrayBuffer` is unavailable, and the browser cannot deliver your input to the game. Any proxy placed in front of it must preserve those headers.
 
 **On a server (`UIType.WEB`)** — the game runs in the Python process and the browser is a terminal for it, with save files on the server's disk under `data/`. Handy for playing over a terminal-less machine on your own network, but everyone who opens the page shares one game and one set of saves:
--- a/src/ui/webUserInterface.py
+++ b/src/ui/webUserInterface.py
@@ -1,3 +1,4 @@
+import errno
 import json
 import os
 import queue
@@ -144,6 +145,30 @@ def __getattr__(name):
     raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
 
 
+def _bindServer(host, port, handler):
+    """Start the game's HTTP server, turning a refused address into a sentence.
+
+    The front-end does not exist yet at this point, so there is no showDialogue
+    to say it through: a port that is already listening - a second copy of the
+    game, most often - would otherwise reach the player as an errno raised from
+    inside http.server, naming neither FishE nor FISHE_WEB_PORT, which is what
+    they would have to change."""
+    try:
+        return ThreadingHTTPServer((host, port), handler)
+    except OSError as e:
+        if e.errno == errno.EADDRINUSE:
+            reason = "something else is already listening there"
+        elif e.errno == errno.EACCES:
+            reason = "this process is not allowed to use that port"
+        else:
+            reason = str(e)
+        raise OSError(
+            f"FishE's web front-end could not be served at http://{host}:{port}/: "
+            f"{reason}. Set FISHE_WEB_PORT to a free port (or FISHE_WEB_HOST to "
+            f"an address this machine can bind) and start the game again."
+        ) from e
+
+
 def _makeRequestHandler(ui):
     """Build a request handler bound to a specific WebUserInterface instance."""
 
@@ -219,7 +244,7 @@ class WebUserInterface(BaseUserInterface):
         self._inputQueue = queue.Queue()
         self._server = None
         if start_server:
-            self._server = ThreadingHTTPServer((host, port), _makeRequestHandler(self))
+            self._server = _bindServer(host, port, _makeRequestHandler(self))
             self._server.daemon_threads = True
             threading.Thread(target=self._server.serve_forever, daemon=True).start()
 
--- a/tests/ui/test_webUserInterface.py
+++ b/tests/ui/test_webUserInterface.py
@@ -6,6 +6,8 @@ import threading
 import urllib.error
 import urllib.request
 
+import pytest
+
 # Use the bare `ui.*`/`player.*` import style (matching production) so class
 # identities line up with the runtime MRO; pytest.ini exposes both `.` and `src`.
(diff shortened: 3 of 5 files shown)


dmccoystephenson:
## Self-review

The full diff was read and checked for logic errors, missed edge cases, front-end-parity gaps, schema/reader-writer mismatches, missing tests on public methods, and doc drift. No blocking issue was found. Three notes are recorded below, along with one gap that has already been closed on this branch.

**Front-end parity** — verified as not applicable rather than skipped. Only `WebUserInterface` and `web/serve.py` bind a socket: `UserInterface`/`ConsoleUserInterface` and `PygameUserInterface` open no server, and `PyodideUserInterface` runs inside the browser tab with no server at all. No `BaseUserInterface` primitive was added, removed, or changed, so nothing here has to be mirrored into the other three front-ends.

**Save-file contract** — no field on `Player`, `Stats`, or `TimeService` was touched, so `schemas/player.json`, `schemas/stats.json`, `schemas/timeService.json` and their `*JsonReaderWriter` classes are correctly left alone.

**Socket cleanup on a failed bind** — checked, because a helper that swallows and re-raises is where a half-open socket would be leaked. `socketserver.TCPServer.__init__` already wraps `server_bind()`/`server_activate()` in a bare `except` that calls `server_close()` before re-raising, so the failed socket is closed before either `_bindServer` ever sees the `OSError`. Nothing further is needed.

### Notes (non-blocking)

- `src/ui/webUserInterface.py:165` — the raised `OSError` is constructed from a message alone, so `errno` and `strerror` are `None` on it. Nothing in the codebase catches this by errno, and `raise ... from e` keeps the original reachable via `__cause__`, so no behaviour is affected. The two-argument form (`OSError(e.errno, message)`) was considered and rejected: it would prefix the player-facing text with `[Errno 98]`, which is the noise this change exists to remove. The same applies at `web/serve.py:118`.
- `src/ui/webUserInterface.py:167` — the message tells the reader to set `FISHE_WEB_PORT`, while `WebUserInterface` also accepts `port` as a constructor argument that owes nothing to the environment (the tests pass it directly). In production the only construction path is `UserInterfaceFactory`'s WEB branch, which does read that variable, so the advice is correct for every player who can see the message; it is only approximate for a caller constructing the class directly.
- `web/serve.py:100` — the `ValueError` is re-raised from inside `except ValueError` without `from None`, so the underlying `int()` error is still shown as context. This is deliberate: `UserInterfaceFactory`'s existing check at `src/ui/userInterfaceFactory.py:46-50` does exactly the same, and the two are meant to behave identically.

### Addressed during review

- `tests/web/test_serve.py` — `_bindServer` was covered only through its failure branch, so an implementation that always raised would have passed. A success-path test asserting that a free port is bound and the server handed back was added in the second commit on this branch.

This comment was drafted during a Gardener session (https://github.com/Stephenson-Software/gardener).

PR #179, opened 2026-08-14
Announce the web app's URL only once its server is bound

### LINKED ISSUE #178: examples/web_app.py announces a URL it has not yet bound, reading FISHE_WEB_PORT a third time to do it
## Summary

`examples/web_app.py:34-38` prints the address before anything is listening on it, and builds that address from its own copy of the two environment variables rather than from the server:

```python
host = os.environ.get("FISHE_WEB_HOST", "127.0.0.1")
port = os.environ.get("FISHE_WEB_PORT", "8000")
print(f"FishE web app is starting at http://{host}:{port}")
print("Open that URL in your browser to play. Press Ctrl+C here to stop.")
game = FishE(interfaceType=UIType.WEB)
```

Two consequences follow, both verified against source:

- **A URL that was never served is printed first.** `FISHE_WEB_PORT=80801x python3 examples/web_app.py` prints `FishE web app is starting at http://127.0.0.1:80801x` and only then raises, because the value is not converted here at all — `UserInterfaceFactory`'s WEB branch (`src/ui/userInterfaceFactory.py:44-50`) does the converting, after the print. The same ordering applies to a port that is already taken, which `WebUserInterface` now reports from `_bindServer` (`src/ui/webUserInterface.py:148-169`) — again after the URL has been announced. `web/serve.py` was given the opposite ordering in #177 (`web/serve.py:128-133`: bound first, announced second), so the two web entry points now disagree about this.
- **The defaults are copied.** `"8000"` is written here and again at `src/ui/userInterfaceFactory.py:44`. Should the factory's default ever move, this file would keep printing the old one, with nothing failing to say so.

The ordering cannot simply be swapped: `FishE.__init__` builds the front-end and then blocks in `_selectSaveFile()` (`src/fishE.py:52-60`), so control does not come back to `main()` after construction, and `WebUserInterface.address` — which would be the authoritative answer — cannot be read from there. Some other arrangement is required, which is why this is filed rather than folded into #177.

## Why it matters

This is the entry point the README points a player at for the server-backed front-end (`README.md:34`). Of the two ordinary startup failures — a misspelled port and a busy one — both are now explained well by the exception, but both are still preceded by a confident sentence naming an address that will never answer.

## Suggested fix

The port could be validated in `main()` before `FishE` is constructed (sharing the check the factory already performs), and the announcement could be deferred until the server is known to be bound — for example by having the factory or `WebUserInterface` report the bound address, or by binding in `main()` and handing the result in. Whichever shape is chosen, the `"8000"` default should be stated in one place rather than two. A test alongside `tests/ui/test_userInterfaceFactory.py:116-128` would cover it.

_Filed by Claude during an automated triage pass; every claim above was verified against source. This issue was filed during a Gardener session (https://github.com/Stephenson-Software/gardener)._

dmccoystephenson:
## Summary

- `examples/web_app.py` printed `http://{host}:{port}` before anything was listening on it, from its own copy of `FISHE_WEB_HOST`/`FISHE_WEB_PORT`. Both ordinary startup failures — a misspelled port and a busy one — were therefore preceded by a confident sentence naming an address that would never answer, since neither the conversion (`UserInterfaceFactory`'s WEB branch) nor the bind (`WebUserInterface._bindServer`) happens until after that print. `web/serve.py` was given the opposite ordering in #177, so the two web entry points disagreed about it.
- The ordering could not simply be swapped in the example: `FishE.__init__` builds the front-end and then blocks in `_selectSaveFile()`, so control never returns to `main()` to say what was bound. The announcement is instead made by `WebUserInterface` immediately after `_bindServer` returns, and names the socket's own address — so a caller that asked for an ephemeral port is told the port it actually got. The entry point now announces nothing of its own, and a failed bind is no longer preceded by a URL.
- The `"8000"` default and the two variables that move it are now stated once, in `webUserInterface` (`DEFAULT_HOST`/`DEFAULT_PORT` and `resolveAddressFromEnvironment()`), and shared by the factory branch and the constructor; the example no longer reads the environment at all. `web/serve.py` keeps its own `8080` default deliberately, as the comment there already explains.
- No behaviour was changed for the Pyodide front-end: it subclasses `WebUserInterface` with `start_server=False`, so it has no address and announces nothing.

## Test plan

- [x] `python3 -m pytest` (with `SDL_VIDEODRIVER`/`SDL_AUDIODRIVER=dummy`) — 829 passed
- [x] New: the bound address is announced with the port the socket actually got
- [x] New: nothing is announced when no server was started, or when the port was already taken
- [x] New: `resolveAddressFromEnvironment()` covers the default, the override, and the misspelled-port `ValueError`
- [x] New `tests/examples/test_web_app.py`: the entry point builds the WEB front-end and prints no address of its own
- [x] `black --check` on the changed files

## Deferred this cycle

No other issue was open at triage time, so nothing was deferred.

Closes #178

---

### DIFF
--- a/examples/web_app.py
+++ b/examples/web_app.py
@@ -14,10 +14,11 @@ Run it and open the printed URL:
     python3 examples/web_app.py
 
 The whole game (save-file manager, fishing, shop, bank, tavern, dialogue) then
-plays in the browser. The server binds 127.0.0.1:8000 by default; set
-FISHE_WEB_HOST/FISHE_WEB_PORT (read by UserInterfaceFactory's WEB branch) to
-change that — e.g. FISHE_WEB_HOST=0.0.0.0 so it's reachable from outside its
-own host, such as from inside a container.
+plays in the browser. The server binds 127.0.0.1:8000 by default (the default
+and the two variables that move it live in WebUserInterface); set
+FISHE_WEB_HOST/FISHE_WEB_PORT to change that — e.g. FISHE_WEB_HOST=0.0.0.0 so
+it's reachable from outside its own host, such as from inside a container. The
+URL is printed once the server is bound, so what it names is always served.
 """
 
 import os
@@ -31,12 +32,12 @@ from fishE import FishE  # noqa: E402
 
 
 def main():
-    host = os.environ.get("FISHE_WEB_HOST", "127.0.0.1")
-    port = os.environ.get("FISHE_WEB_PORT", "8000")
-    print(f"FishE web app is starting at http://{host}:{port}")
-    print("Open that URL in your browser to play. Press Ctrl+C here to stop.")
     # Building FishE starts the web server and then waits (in the save-file
     # manager) for the browser to interact, so play happens entirely in-browser.
+    # The URL is announced by the front-end once that server is bound, rather
+    # than from here beforehand: control never comes back to this function to
+    # say what was bound, and a port that is misspelled or already taken would
+    # otherwise be preceded by an address that will never answer.
     game = FishE(interfaceType=UIType.WEB)
     game.play()
 
--- a/src/ui/userInterfaceFactory.py
+++ b/src/ui/userInterfaceFactory.py
@@ -1,5 +1,3 @@
-import os
-
 from ui.enum.uiType import UIType
 from ui.consoleUserInterface import ConsoleUserInterface
 from prompt.prompt import Prompt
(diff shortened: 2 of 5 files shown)


dmccoystephenson:
Self-review rubric (scored against the diff and command output, not judgement):

- **Scope: PASS** — four files are touched plus one new test file, all of them named by #178: the example that announced too early, the factory branch that held the duplicate default, the front-end that binds, and their tests. No unrelated reformatting is present (`black --check` reports the five changed files unchanged, so nothing was reformatted into the diff).
- **Tests-new: PASS** — the one new public function, `resolveAddressFromEnvironment()`, is exercised three ways (default, override, misspelled port); the new private `_announceAddress()` is exercised through the constructor by three tests.
- **Tests-fix: PASS — empirically confirmed.** With `src/` and `examples/` reverted to `origin/main` and the tests kept, exactly the five new tests fail (`5 failed, 37 passed`); restored, `42 passed`. The failure is observable rather than incidental because the announcement is asserted to name the *ephemeral* port the socket actually got, which the reverted code could not have printed.
- **Sibling structure: PASS** — `tests/examples/test_web_app.py` follows the shape of `tests/web/test_serve.py` (module imported by package path, no `__init__.py`, `# check` comments, `unittest.mock` for collaborators).
- **Sibling renames: PASS** — no identifier in a parallel pair or series was renamed; `_bindServer`/`address`/`cleanup` keep their names.
- **Docs: PASS** — `README.md:34-36` still describes what happens (`python3 examples/web_app.py`, then `http://127.0.0.1:8000`), only now that URL is printed after the bind rather than before it; the example's module docstring was updated to say where the default lives and when the URL is printed. No schema, `PLANNING.md` goal, or persisted field is affected by this change.
- **Issue resolution: PASS** — both halves of #178 are addressed: the announcement moved behind the bind, and the `"8000"` default is stated once.
- **CI: PASS** — `test` passed in 39s on the PR head; `python3 -m pytest` under the dummy SDL drivers reports 829 passed locally.
- **Schema-sync: N/A** — no persisted field is added or changed, so no `schemas/*.json` or `*JsonReaderWriter` is touched.
- **Money-format: N/A** — no money or price is displayed by this diff.
- **Deterministic-tests: PASS** — no new test uses `random`; the two that bind sockets ask for port 0 and read back the port that was assigned, rather than assuming one.
- **Headless-pygame: N/A** — no pygame code path is touched; the web front-end's servers bind loopback and are closed in `finally`.
- **camelCase: PASS** — `resolveAddressFromEnvironment`, `_announceAddress`, `boundHost`, `boundPort`, `portText` follow the surrounding style. (`host`/`port` keyword arguments keep the names the constructor already had.)

Two observations that were judged rather than measured, folded in here rather than left inline:

- `src/ui/webUserInterface.py:256` — the announcement is unconditional whenever a server is started, so there is no way for an embedder to silence it. No `announce=False` knob was added because no caller wants one: the only two constructors are the factory (which serves a player) and the Pyodide subclass (which passes `start_server=False` and so prints nothing). Should a quiet mode ever be wanted, this is where it would go.
- `examples/web_app.py:17` — the module docstring still states `127.0.0.1:8000` in prose, which is a third mention of the number after `DEFAULT_HOST`/`DEFAULT_PORT`. It was kept because a reader of the example benefits from seeing the address without following it into `webUserInterface`, and it now says where the authoritative copy lives; only the *code* copies were removed.

This comment was drafted during a Gardener session (https://github.com/Stephenson-Software/gardener).

---

